PR #9, opened 2018-09-04
fix: check for window to fix Gatsby builds

jlengstorf:
This adds a `typeof` check to `window` and adds a helper function to abstract that logic to a single place. This will allow Gatsby to build because it short-circuits the `window` access during the SSR step.

### DIFF
--- a/src/components/SideNav/SideNavItem.js
+++ b/src/components/SideNav/SideNavItem.js
@@ -3,6 +3,9 @@ import classnames from 'classnames';
 import { Link } from "gatsby";
 import { Icon } from 'carbon-components-react';
 
+const getPathArray = () =>
+  typeof window !== 'undefined' && window.location.pathname.split('/');
+
 export default class SideNavItem extends React.Component {
 
   state = {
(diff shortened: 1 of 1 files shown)


src/components/SideNav/SideNavItem.js
```diff
@@ -3,6 +3,9 @@ import classnames from 'classnames';
 import { Link } from "gatsby";
 import { Icon } from 'carbon-components-react';
 
+const getPathArray = () =>
+  typeof window !== 'undefined' && window.location.pathname.split('/');
```
asudoh: Oops I missed array de-ref of `getPathArray()` callers - Do you want to return an empty array in SSR scenario, or let caller check for `undefined`, to avoid NPE?

jlengstorf: Good call — this could probably be improved further. The goal is to get a boolean back for `classnames`, so maybe there's a simpler shortcut.

```js
const isBrowser = typeof window !== 'undefined';
const pathArray = isBrowser ? window.location.pathname.split('/') : [];

// ...

const isSubActive = isBrowser && pathArray[2] === item;
const subNavClasses = classnames('side-nav__sub-nav-item', {
  'side-nav__sub-nav-item--active': isSubActive,
  'side-nav__sub-nav-item--active': isSubActive,
});
```

That should avoid errors, and has an added benefit of being a bit more descriptive.

asudoh: The code you brought up makes much more sense - Thanks @jlengstorf!

PR #14, opened 2018-09-10
feat(ComponentExample): instantiate vanilla components

asudoh:
Brought (back) in some code from the original Carbon site code.

I initially tried the approach of using `CDS` global like current Carbon site code does, but took an alternate approach (using equivalent ESM import) as Gatsby ignores our WebPack config that create `CDS` global.


alisonjoseph:
Hi @asudoh the components still don't appear to be working? Is there something else I need to add/update? thanks for taking a look at this for me!

Issue #24, opened 2018-09-13
IE11 site issues


alisonjoseph:
@ASUDOH the pollyfills you added break the gatsby build because you can't use `window....` with gatsby 😢 

<img width="557" alt="screen shot 2018-09-13 at 8 08 02 am" src="https://media.github.ibm.com/user/1823/files/323b7496-b72c-11e8-991d-e627728bdce1">

https://www.gatsbyjs.org/docs/debugging-html-builds/

I wonder if some of the docs here would help instead? 
https://next.gatsbyjs.org/docs/browser-support/#browser-support

Currently the test version of the site is here https://carbon-design-system.github.io/carbon-website-gatsby/ but it is not working correctly in IE11. 

any thoughts?

alisonjoseph:
@asudoh just checked the production build of the site and it is working in IE11 🎉 (you can view here https://carbon-design-system.github.io/carbon-website-gatsby/)

PR #36, opened 2018-09-14
fix(config): IE11 fixes

asudoh:
Refs #24.

Does a few things:

1. Changed the `excludes` option of `babel-loader`. [`ansi-regex`](https://www.npmjs.com/package/ansi-regex) module seems to be loaded in this app. That module has `const` which cannot be parsed by IE11 and thus needs to be transpiled.
2. A change as an attempt to load our polyfills at the top-level
3. Check for web platform globals to cope with Gatsby SSR

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -84,12 +84,25 @@ exports.createPages = ({ actions, graphql }) => {
   });
 };
 
-exports.onCreateWebpackConfig = ({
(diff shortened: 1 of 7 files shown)


asudoh:
@alisonjoseph I think we can merge - The remaining issue does not have to do with this change. Thanks!

Issue #27, opened 2018-09-14
Experimental Component Styles inside Gatsby site


alisonjoseph:
@asudoh still seeing some styling inconsistencies. 

Thanks so much for all your help on this! 

Gatsby Site
<img width="233" alt="screen shot 2018-09-13 at 8 27 58 pm" src="https://user-images.githubusercontent.com/2753488/45524369-b7693100-b793-11e8-9fb6-b33969277397.png">

vs. the dev env.
<img width="246" alt="screen shot 2018-09-13 at 8 28 03 pm" src="https://user-images.githubusercontent.com/2753488/45524367-b46e4080-b793-11e8-9f24-23c555330e0f.png">

Also, when I do a production build the styles don't appear at all.  Latest prod build -> https://carbon-design-system.github.io/carbon-website-gatsby/

asudoh:
Hi @alisonjoseph do you have a quick instruction on how to create the production build? Thanks!

alisonjoseph:
Sorry when I said dev env I meant `carbon-components` repo with experimental turned on, I'm seeing the same styles on gatsby dev and prod now. You can see that some of the styles for checkbox are not coming over for some reason. 

So this is what checkbox should look like and what it looks like in carbon-components
<img width="246" alt="screen shot 2018-09-13 at 8 28 03 pm" src="https://user-images.githubusercontent.com/2753488/45524367-b46e4080-b793-11e8-9f24-23c555330e0f.png">

and this is what is displaying on the Gatsby Site (both in dev and prod builds)
<img width="233" alt="screen shot 2018-09-13 at 8 27 58 pm" src="https://user-images.githubusercontent.com/2753488/45524369-b7693100-b793-11e8-9fb6-b33969277397.png">

Here is an example comparison of the focus styles with the DOM inspector between carbon-components and gatsby site.
<img width="992" alt="screen shot 2018-09-14 at 7 59 55 am" src="https://user-images.githubusercontent.com/2753488/45551650-41e17d00-b7f4-11e8-9e91-e047d2d2d3dd.png">
<img width="690" alt="screen shot 2018-09-14 at 7 59 31 am" src="https://user-images.githubusercontent.com/2753488/45551651-41e17d00-b7f4-11e8-9e9b-d9b646b3b1c0.png">

PR #37, opened 2018-09-14
fix(styles): separate theme override

asudoh:
Refs #27.

Checkbox position is off with this change somehow - @alisonjoseph great if you have any ideas, thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@reach/router": "^1.1.1",
     "@storybook/addon-info": "^3.4.7",
-    "carbon-components": "^9.6.1",
+    "carbon-components": "^9.12.0",
     "carbon-components-react": "6.18.3",
     "carbon-icons": "^7.0.7",
     "classnames": "^2.2.6",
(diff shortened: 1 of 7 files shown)


alisonjoseph:
This is closer, thank you. What styles are you noticing that are off? the only thing Im seeing is this focus style that shouldn't be there. 
screen shot 2018-09-14 at 11 32 05 am

I wonder if we could use a different prefix for the experimental styles and swap the html also so that the styles don't interfere with each other?

asudoh:
Thank you @alisonjoseph for testing! The root of the issue is that the only ways to fully scope styles are either `<iframe>` or [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot). The former makes UX worse, and the latter is not supported by IE/Edge.

That said, are you suggesting we use a completely different page for experimental vs. stable? If it can be done with Gatsby and you think it’s the way for us to go, I’m all for it! 😉

alisonjoseph:
I think it might be better to keep the code sample as `bx--` because when someone actually goes to use the experimental components they most likely won't be changing the prefix.

asudoh:
Got it @alisonjoseph! Updated the code as such. The prefix is an idea that I just rattled off, but feel free to suggest any alternatives!

PR #47, opened 2018-09-17
fix(devenv): patch WebPack entry config

### LINKED ISSUE #24: IE11 site issues

asudoh:
This change ensures `Symbol` polyfill is loaded before `react` module is loaded. Doing so workarounds gatsbyjs/gatsby#7003, which happens because `Symbol` polyfill is loaded after `react`and before `react-dom`, which makes `react` and `react-dom` refer to different `Symbol` definitions.

Fixes #24.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -84,12 +84,33 @@ exports.createPages = ({ actions, graphql }) => {
   });
 };
 
-exports.onCreateWebpackConfig = ({ actions, getConfig }) => {
+exports.onCreateWebpackConfig = ({ actions, getConfig, stage }) => {
   const config = getConfig();
-  const { module } = config;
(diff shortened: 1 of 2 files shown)


alisonjoseph:
Thanks Akira!

PR #54, opened 2018-09-18
Add DocSearch

### LINKED ISSUE #17: Create Search Component
https://codepen.io/alisonjoseph/pen/yxRqdX?editors=1010

Carbon config file
https://github.com/algolia/docsearch-configs/blob/master/configs/carbondesignsystem.json
PR to config: https://github.com/algolia/docsearch-configs/pull/563

alisonjoseph:
closes #17 

Needs some design input, this is just default style. fyi: @designertyler @claycrenshaw 

Can view in codepen here https://codepen.io/alisonjoseph/pen/yxRqdX?editors=1010

Screen shot on the website, I added the same layer styles as overflow menu and updated all the font sizes/colors/padding/hover to match our tokens. 

screen shot 2018-09-18 at 11 08 29 am


src/components/GlobalSearch/GlobalSearch.js
```diff
@@ -0,0 +1,44 @@
...
+          className="bx--search-magnifier"
+        />
+        <label htmlFor="doc-search" className="bx--label">
+          Search
+        </label>
+        <input
+          aria-label="Search"
+          type="text"
+          className="bx--search-input"
+          id="doc-search"
```
asudoh: Is this a singleton component? If not, we need a way to avoid ID duplication, e.g.: https://github.com/IBM/carbon-components-react/blob/v6.32.1/src/components/PaginationV2/PaginationV2.js#L259

alisonjoseph: This is the global search for the website, it would never be used more than once.

claycrenshaw:
Alison, this is an awesome POC! Thank you for pulling this together so fast! Is the algorithm the same as on the current site? We definitely need some design work on the interface. . . But this is a great start :)

alisonjoseph:
We are using DocSearch for this, they auto-index the site every 24 hours, so this is better than the current site, and much less work on our part.

jeoffw:
👏 Awesome!!

p.s. I like Algolia.

PR #62, opened 2018-09-19
chore: use custom .babelrc

asudoh:
Gatsby's default `.babelrc` is configured in a manner that automatically imports Babel polyfills
based on usage of corresponding API.

Such Babel polyfills are loaded after RHL imports `react` module and before page bundle generated by Gatsby imports `react-dom` module, which causes `react` and `react-dom` referring to different `Symbol` class object, triggering gatsbyjs/gatsby#7003.

Refs #24.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,40 @@
+{
+  presets: [
+    [
+      "@babel/preset-env",
+      {
+        loose: true,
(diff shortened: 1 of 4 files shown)


alisonjoseph:
Thanks for researching this! 🙌

PR #85, opened 2018-09-21
ci: add CircleCI support

joshblack:
Adds staging site support through CircleCI. The idea is that any change that is merged into master is continuously deployed to http://staging.carbondesignsystem.com 👍 

For the setup, we use an offline mirror of Yarn to verify that we are deploying with stable (and hopefully secure) dependencies. In addition, we install the `ibmcloud` CLI to handle deployments. 

Using an API Key from IBM Cloud, we can login to IBM Cloud and use the `cf` (Cloud foundry) command to do a Staticfile deployment. The Staticfile deployment basically uploads the `public` folder after it is built and then serves it using NGINX under the hood.

We use IBM Cloud's routing feature to map the deployment to staging.carbondesignsystem.com.

**New**

- `.circleci`
- `.cfignore`
- `Staticfile`

**Changed**

- Update paths to be case sensitive. On MacOS `components/layouts` could resolve to `components/Layouts`, but it would fail on Linux machines

**Removed**

### DIFF
--- a/.circleci/manifest.staging.yml
+++ b/.circleci/manifest.staging.yml
@@ -3,6 +3,5 @@ applications:
 - name: carbon-website-staging
   memory: 64M
   buildpack: https://github.com/cloudfoundry/staticfile-buildpack.git
-  random-route: true
-  # routes:
-  # - route: staging.carbondesignsystem.com
+  routes:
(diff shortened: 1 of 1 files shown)


joshblack:
@alisonjoseph should I also try and add support for the internal site deploy?

Issue #43, opened 2018-09-14
Create content describing and pointing to "Experimental" components and functionality


shinytoyrobots:
A user needs to discover and understand what experimental components (and other resources) are available and how to use them.

DESIGN
- [ ] replicate the existing method Carbon site uses to show components, for subset of experimental components

WEBSITE
- [X] [Explanation of what "Experimental" is](https://github.com/carbon-design-system/carbon-website-gatsby/pull/38)
- [X] design IA for discovering experimental components
- [ ] Explanation of how to turn on a feature flag, in order to use/test components (related to https://github.com/IBM/carbon-components-react/issues/1339) http://staging.carbondesignsystem.com/experimental/about

PR #111, opened 2018-09-25
feat: add docs for exp-component feature-flag usage

chrsfshr:
First pass at adding docs for feature-flag usage for experimental-components

Issue reference: #43


alisonjoseph:
Hey @chrsfshr this is a good start! 

Users that would need to turn on a feature flag wouldn't actually touch any of our sass files, they would be pulling in `carbon-components` using npm. They would then add this to their own sass file in their project and turn on the flag that they need.

``
$feature-flags: (
  components-x: false
);
``

chrsfshr:
Oh wow ok thats simple enough!...so they would just add this to their own local file then run it.

alisonjoseph:
Yep, might be helpful to include how they can import carbon-components in and then add the feature flag. 

So their file would look something like this

```
.carbon-demo-experimental {
  $feature-flags: (
    components-x: true,
    ui-shell: true,
  );
  @import 'carbon-components/src/globals/scss/styles';
```

alisonjoseph:
We are actually using this in the Gatsby site to display the experimental components. You can see the file setup here https://github.com/carbon-design-system/carbon-website-gatsby/blob/master/src/styles/experimental.scss We're also using the custom prefix variable and wrapping everything in a container class.

alisonjoseph:
Should probably also add that they must be using sass and not the compiled css file in order to take advantage of the code using feature flags.

alisonjoseph:
We should also add some docs for how to turn it on for the react library and/or link to the docs once this PR is merged in https://github.com/IBM/carbon-components-react/pull/1347

chrsfshr:
Seems pretty slim content wise...Not sure If I'm missing a step

src/content/experimental/about/usage.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+**Note: You must be using sass and not the compiled css file in order to take advantage of the code using feature flags.** 
+
+
+
+## Disclaimer 
+
+Experimental components, designs, patterns, and other resources are work that is not fully tested. But they are in a state where it can be productive for designers and developers to work with them.
+
```
alisonjoseph: Can we maybe re-word this paragraph so we don't have a sentence starting with But.

alisonjoseph:
Should we add something to explain that we have currently have a feature flag for the experimental components and also a separate one for the ui-shell?

src/content/experimental/about/usage.md
asudoh: ^ 👍 - Another thought I have here is "not fully tested" wording, because every change should go though testing in some ways. I assume this means "UX testing" or a like?

asudoh:
^ @alisonjoseph I think it makes sense!

src/content/experimental/about/usage.md
alisonjoseph: Good point, lets work on re-wording this

PR #141, opened 2018-09-27
ci: confirm site can build in CI

joshblack:
Adds step to CI to confirm site can build with changes. Should require status checks to go green before merging 👍


.circleci/config.yml
```diff
@@ -9,13 +9,15 @@ jobs:
       - run:
           name: Install dependencies from offline mirror
           command: yarn install --offline
+      - run:
+          name: Confirm project can build
+          command: |
+            yarn build:external
+            yarn build:internal
```
alisonjoseph: If it runs both of these and then deploys, will it then deploy the internal build? Sorry not 100% familiar with how this works. 😄

joshblack: Good call! swapping

PR #155, opened 2018-09-29
Update component status and component overview

### LINKED ISSUE #153: Remove UI Shell from Component Overview Page

alisonjoseph:
Closes #153

### DIFF
--- a/src/components/ComponentStatus/ComponentStatus.js
+++ b/src/components/ComponentStatus/ComponentStatus.js
@@ -16,11 +16,11 @@ class ComponentStatus extends React.Component {
     let react;
     let angular;
     let experimental;
-    if (currentItem.status === 0) {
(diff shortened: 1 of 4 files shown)


src/components/ComponentStatus/ComponentStatus.js
```diff
@@ -77,8 +77,6 @@ class ComponentStatus extends React.Component {
         <td>
           {currentItem.item} {tag}
         </td>
-        <td>{currentItem.added}</td>
```
asudoh: Just wanted to double-check if removal here is intentional.

alisonjoseph: Yes, was removing this column since we are showing vanilla, react, angular and experimental the version things were added didn't seem as relevant.

PR #162, opened 2018-10-01
Cleanup of component page content and images

### LINKED ISSUE #138: Gray background for component images is missing
The previous website would use transparent backgrounds for component images and the grey background was generated by code. This means every png used for components had a transparent background. [Example png](http://www.carbondesignsystem.com/images/modal-usage-1.png) and [The png on page with a grey background](http://www.carbondesignsystem.com/components/modal/usage)

Is this something we want to cary over to the new Gatsby site? What was the reason to do this on the previous site?

alisonjoseph:
Closes #138 

Adds light blue background color to all images on component pages. 

Added a classname to any pages with the `label: Component` in the title. So this won't work on experimental images, which should be fine since they wouldn't have the blue background anyway. 

Also some general cleanup of markdown files with stuff that was missed when copy/pasting over from old site.


src/templates/page.js
```diff
@@ -80,6 +81,12 @@ export default ({ data, pageContent }) => {
   const { GATSBY_CARBON_ENV } = process.env;
   const isInternal = GATSBY_CARBON_ENV !== 'internal' && internal == true;
 
+  const classNames = classnames('page-content', {
+    'page-content--component': post.frontmatter.label === 'Component'
+  });
+
+  console.log(post.frontmatter.label);
```
asudoh: @alisonjoseph Was this a non-temporary debugging output?

alisonjoseph: oops! will fix

PR #178, opened 2018-10-02
Updates to exp. Color in UI page

claycrenshaw:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}


claycrenshaw:
@alisonjoseph I considered using that style, but I think the bigger color blocks and this organization offers a little more to the user. As for the <color-block> component, I'll look into that for the next iteration!

alisonjoseph:
If we need to update a component based on a new layout we can do that, or add a prop for a larger color block.

alisonjoseph:
going to merge this in for now so its up on staging. but should definitely discuss best way to match this page to our current components/styles 😄

PR #189, opened 2018-10-03
159 mobile nav flickers open closed

dakahn:
This is a super temporary fix to squash this bug using an old-school lifecycle method so that we check the page width _before_ the component is rendered rather than after.

Moving forward the nav logic will need to be reworked to get componentDidMount() working as intended without the nav popping in and out.


alisonjoseph:
This actually causes a build error with window. going to revert 😢

PR #218, opened 2018-10-05
159 mobile nav flickers open closed

### LINKED ISSUE #159: Mobile nav flickers open/closed when you click on tabs
@dakahn fyi. let me know if you need more details.

dakahn:
closes #159 . 

Trouble is, the ghpages build of my fork weirdly redirects to test1.carbondesignsystem (which is a 404 :hurtrealbad:). So whoever checks this out do me a solid and test that the _built_ site no longer has the nav pop in and out in a refreshed mobile view when clicking through nested routes (like in Contributing, for example).


lovemecomputer:
when i got to `http://localhost:8000/contributing/designers` and refresh the page it still starts open and immediately closes :(

dakahn:
From a refreshed mobile view? Dev works in both FF and Chrome for me. :(

oct-05-2018 12-54-57

But if the user story for this bug is that the user starts in desktop and then shrinks their browser _down to a mobile view_ -- that's definitely do-able, but a little more intensive.

alisonjoseph:
So this definitely works if you start at mobile size, but if you resize from small to big or big to small weird stuff happens.

lovemecomputer:
@dakahn @alisonjoseph oh yea it totally works as that gif illustrates!! i meant actually hitting the "refresh" button like the original post described, which is probably edge-case/nothing to worry about. works great for me when navigating between pages!

alisonjoseph:
closing this for now. we can re-look at this bug in the future.

PR #220, opened 2018-10-05
fix(SideNavItem): use starts-with matching for location/nav

### LINKED ISSUE #209: Active class doesn't show up on pages w/ tabs
For example http://www.carbondesignsystem.com/getting-started/developers/vanilla

Developers should be highlighted in the left nav

asudoh:
Earlier PR #176 seems to allude that we need to detect if navitem/subnavitem/tab URL starts with navitem/subnavitem from nav data to see if a sub-nav-item should be activated, and this change ensures that we do such comparison.

Fixes #209.

#### Changelog

**Changed**

- `locationContainsPath()`, merged with `locationContainsPathAtIndex()`, to do "starts-with" search of path tokens, instead of "a token at somewhere" or "a token at specific location" search.
- `normalizePath()` (renamed from `normalizeLocation()`) to accept array of strings


alisonjoseph:
I don't think we can remove the paths from the nav.

We added them because you weren't able to navigate to "http://carbondesignsystem.com/guidelines/content/guidance" before, you would get a 404 because the redirect doesn't work correctly.

asudoh:
@alisonjoseph My apologies for the oversight - Brought back the paths for now, though we lose the nav item active state for http://www.carbondesignsystem.com/getting-started/developers/react. (http://www.carbondesignsystem.com/getting-started/developers/vanilla works with this fix)

PR #226, opened 2018-10-05
feat(surge): add easy surge deploy

tw15egan:
Adds in easy `Surge` deploy steps (https://surge.sh/)

A user would simply need to run `yarn deploy` from the command line to publish a working example of their site.


package.json
```diff
@@ -23,7 +23,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "storybook:build": "build-storybook -c .storybook -o public/docs",
-    "deploy": "gatsby build && gh-pages -d public"
+    "deploy": "yarn build:external && cd public && surge"
```
alisonjoseph: We should add in options for internal and external, I think most of the content the designers are working on is internal. 

so something like
`deploy:internal": "yarn build:nternal && cd public && surge"`

tw15egan: Good point. I'll update the PR to include two options

PR #231, opened 2018-10-06
fix(build): create index.html for sub-nav items with tabs

asudoh:
(Please use Diff settings - Hide whitespace changes for review)

Debugging how `redirects.json` (the file created by Gatsby's [`createRedirect()` API](https://www.gatsbyjs.org/docs/actions/#createRedirect)) as well as issue 5329 at gatsbyjs/gatsby and Gatsby author's comments at issue 2511 gave me an impression that we need `.html` file representing from-redirect path so Gatsby's client-side redirect feature to work in static file build, and it seems the case as adding such `.html` file fixed the redirects, e.g. from `/getting-started/developers` to `/getting-started/developers/vanilla`.

This change also fixes an issue where redundant items were created in `redirects.json`.

Refs #176, #209.

#### Changelog

**New**

- Static file build `.index.html` for sub-nav items with tabs, e.g. `/getting-started/developers/index.html`.

**Changed**

- Fix for an issue where redundant items were created in `redirects.json`.

**Removed**

- The first tab from the sub-nav keys in `navigation.json`.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -36,44 +36,53 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 exports.createPages = ({ actions, graphql }) => {
   const { createRedirect, createPage } = actions;
 
-  return new Promise((resolve, reject) => {
-    graphql(`
-      {
-        allMarkdownRemark {
-          edges {
-            node {
-              fields {
-                slug
-                currentPage
-              }
-              frontmatter {
-                tabs
-              }
+  return graphql(`
+    {
+      allMarkdownRemark {
+        edges {
+          node {
+            fields {
+              slug
+              currentPage
+            }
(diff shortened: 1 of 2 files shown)


joshblack:
Was thinking about this over the weekend, could we do a custom `nginx` buildpack that basically tries to serve the file at the path and then falls back to just distributing our index bundle? Then we could have client-side route kick-in (I think). Could also help with deployments so that we could write nginx rules for `/design/product` without having to change our build folder.

asudoh:
@alisonjoseph Yes.

@joshblack I thought about that option, but I also see team's plan to use GH pages in some way, IIRC. I took a route without a backend-dependency therefore, but feel free to choose an alternate approach to solve this problem.

PR #237, opened 2018-10-08
fix: set font to plex for experimental components

alisonjoseph:
experimental components were displaying Helvetica instead of Plex

I'm not entirely sure WHY we have to set this to true. I wonder if there are any other global variables we will need to set for the website?


asudoh:
@alisonjoseph Your guess is correct, it's because: https://github.com/IBM/carbon-components/blob/v9.27.0/src/globals/scss/_css--font-face.scss#L63 - That said, is it possible for you to add a comment saying: `Dummy value so `global-variable-exists()` check in `@include exports('css--font-face')` work`? Thanks!

PR #243, opened 2018-10-10
Image bugs

### LINKED ISSUE #239: Broken images in component pages
There are several broken images on multiple component pages (mostly the style pages).
I'm not sure if it's formatting or the images have been moved

alisonjoseph:
Closes #239

Many of the images on component pages were using a hack/workaround with markup for styling or a custom component for a grid layout. These were missed when copying over content to the new website.

#### Changelog

**Changed**

Went through page by page to try and catch all images that weren't styled correctly, or were displaying too large or too small. 

Added docs around when to use `image-grid` and `image-component` wrapper divs. 

Added image-grid css from old site back in.

**Testing**
Make sure nothing is broken.


alisonjoseph:
merging this fix in. this should hopefully hit 95% of all of the broken/mis-sized component images. We should do another pass/check at some point though. cc @designertyler

PR #260, opened 2018-10-12
Make side nav keyboard accessible.

### LINKED ISSUE #258: Keyboard Accessibility is Broken
Unable to tab through from left navigation to tab on the right. Was working in previous versions.

alisonjoseph:
Closes #258 

#### Changelog

**New**

- Adds tab index to side navigation to make sure its keyboard accessible
- Adds skip to content link

**Testing**

Test side nav with keyboard and make sure it works

cc @bocampbell

### DIFF
--- a/src/components/SideNav/SideNavItem.js
+++ b/src/components/SideNav/SideNavItem.js
@@ -69,8 +69,6 @@ export default class SideNavItem extends React.Component {
         return '';
       }
 
-      console.log(this.state.open);
-
       const tabIndex = this.state.open;
 
       return (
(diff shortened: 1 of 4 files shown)


src/components/SideNav/SideNavItem.js
```diff
@@ -69,11 +69,16 @@ export default class SideNavItem extends React.Component {
         return '';
       }
 
+      console.log(this.state.open);
```
asudoh: Did you mean to remove this line...?

PR #308, opened 2018-10-22
fix(overview-page-list): improve responsive styling for two column layout

### LINKED ISSUE #200: At 935px window width, 2 column makes text overflow when menu opens
image

samrose3:
Closes #200

Prefer CSS grid over flexbox to improve  two column layout responsiveness on the [Overview](https://www.carbondesignsystem.com/) page.

### Changelog

**Changed**

- Convert `display: flex` to `display: grid` on the `.overview-page__list` element
- Adjust children CSS accordingly to accommodate changed display style

**Removed**

- Several media queries that where specific to the children of the `.overview-page__list`

### Screenshots

| Before | After |
| --- | --- |
| before | after |


samrose3:
@alisonjoseph could we consider adding the `gatsby-plugin-postcss`?

In addition, the [`postcss-preset-env`](https://github.com/csstools/postcss-preset-env) plugin would convert modern CSS to a supported equivalent, determining the polyfills as needed based on the targeted browsers.

This might be advantageous for future development on the site as well. Gives the developers a bit more freedom to write cleaner, more modern code 😅

asudoh:
Thanks @samrose3 for sharing your thought! Probably others can chime in, but AFAIK IE11 implementation of CSS grid is older (different) spec than one implemented in modern browsers. Therefore I think we need closer look to see if IE11 works well, even after it goes through autoprefixer. That said, if you can share what you have tested here, it'll be great. Thanks!

alisonjoseph:
Agree w/ Akira, css grid behaves differently in IE11 so a simple pollyfill I don't believe will work in all cases. Have you tested?

samrose3:
I have not. I do agree though that since the implementation is different, testing would be needed. I can revise to an all flex implementation or maybe provide a fallback if grid is not supported

alisonjoseph:
Going to close this due to inactivity. We are releasing a new design/layout next month anyway.

PR #324, opened 2018-10-25
Fix link for tabs with multiple spaces

### LINKED ISSUE #323: Page tabs should support multiple spaces
`tabs: ['Library', 'Usage', 'Design and Contribution']`

If a tab name has more than one space it isn't replaced in the link with a `-` which causes the link to break. Also should not capitalize every since world.

alisonjoseph:
Closes #323

Updates so correct links are added for tabs w/ multiple spaces, and gets rid of css to capitalize all words.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -72,7 +72,7 @@ exports.createPages = ({ actions, graphql }) => {
         },
       });
       if (tabs.length > 1) {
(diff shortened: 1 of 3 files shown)


alisonjoseph:
Good question, going to merge this in since it is blocking Clay's content PRs. I don't see us ever having two spaces like `foo  bar bar`

PR #366, opened 2018-11-07
Add articles & featured back to website

alisonjoseph:
Looks like the articles and featured links got removed when moving FAQ. This adds them to the bottom of the Resources page for now. @claycrenshaw @designertyler is this ok as a temporary place for them? I know long term we want the latest on the homepage and then a list somewhere else?


designertyler:
Agree this is better for now until we find a more permanent place for this content.

claycrenshaw:
yep, makes sense to me.

PR #371, opened 2018-11-07
WIP: Pagination: Next & Previous buttons on content pages

lovemecomputer:
This is a WIP PR! There are a few complexities that will be good to talk through or get opinions on from the team. _(Note: the pagination buttons are currently displayed at the top of page content for easy testing. on final implementation, they will be aligned up against the footer)_

This PR implements the a new `<Pagination>` component, and adds this new component to `src/templates/page.js`

#### Currently, the `<Pagination>` component needs:
- [ ] Solution for linking to a specific tab (tabs on each page is not currently stored in `navigation.js`
- [ ] check if a link target is internal, and skip it from pagination if it shouldn't be displayed.
- [ ] Properly get the `title` attribute from the page data (right now, it's converting the url slug to capitalized text, a quick fix while working out these other issues)

The current state, the code in `Pagination.js` is a little complex, in part because of how the navigation is currently implemented in the site.

#### Info on current form of site navigation data:
- there are two possible levels of hierarchy, `section` and `sub-nav`. There is always the top level `section`, but only sometimes a `sub-nav`. Either of those levels can be a rendered page, and the page(s) markdown files is what stores the tab data in the front matter
- currently, data about tabs live in the markdown files only.
- with the addition of these components, there will now be 3 places where we check for the `internal` flag: `components/SideNav/SideNavItem.js`, `Page.js`, and this new `Pagination` component.
  - should we assign this functionality to a global context/component, so that we are not 

#### So, I guess a big question is:
to get proper link targets in pagination across multiple levels of hierarchy, should we:
1) add additional functionality to `Pagination.js` to extract data on the tabs from potential link target pages's markdown files?
2) refactor navigation data on the site, and include the tabs in `navigation.json`?

Note that change 2 will affect how markdown files are written. The current implementation makes it easy for an author to add pages by interacting with the markdown file(s) only. 

Either way, it's important to set good precedence for use of this new pagination component!

**After meeting, additional solution:**

- Implement the pagination links to only walk through next or previous tabs on the current page. this would make the implementation much simpler (this functionality already works!).
  - additional links could be added that lead to another manually-chosen section! 

(all of the potential solutions come with design and content questions)


lovemecomputer:
closing this in favor of https://github.com/carbon-design-system/carbon-website/compare/master...lovemecomputer:next-previous?expand=1 so that it works with all the new v10 styles!

PR #384, opened 2018-11-12
Add experimental overflow menu

alisonjoseph:
Adds experimental overflow menu to the website (& adds in text field updates w/ latest component version)

Need to pull in experimental js for overflow menu 🤔


alisonjoseph:
@asudoh is the positioning of this off because we aren't getting this? Wasn't sure what else to add to get this example working.
https://github.com/IBM/carbon-components/blob/master/src/components/overflow-menu/overflow-menu.js#L65

asudoh:
Hi @alisonjoseph we may need to add some logic to https://github.com/carbon-design-system/carbon-website/blob/26eb7c5/src/components/ComponentExample/ComponentExample.js#L120-L131 depending on what `direction` is. It calculates the difference between top/left position of the viewport (what the original overflow menu code calculates) vs. top/left position of the `div.component-example__live` (the supposed position).

PR #397, opened 2018-11-14
content cleanup

claycrenshaw:
General content cleanup for getting started:designers


src/content/getting-started/designers/designers.md
```diff
@@ -47,17 +45,16 @@ The Carbon Design System team often makes improvements to the Design Kit. We tra
 	- To update to the latest Carbon library version you will have to download the newest version of the Carbon Design Kit each time it is released and replace the previous version.
 	- Follow the specific [instructions](https://github.com/IBM/carbon-design-kit/wiki/Sketch-Libraries-Overview) on the wiki to ensure a smooth transition.
 
-## Use with Box Drive
+### Sync via Box Drive
```
aagonzales: Not sure we should say "Sync" since the old Box application is `Box Sync` which we don't want people to use. Maybe "connect"?

src/content/getting-started/designers/designers.md
```diff
@@ -47,17 +45,16 @@ The Carbon Design System team often makes improvements to the Design Kit. We tra
...
+_For IBM internal users only._ Using Box Drive sync will ensure that your design kit file will always be up to date, eliminating the need to manually download new versions as they are released.
 
-1. Follow the instructions on [IBM Digital Design](https://www.ibm.com/standards/web/design-kit/) to get the IBM Design kit through Box Drive.
-- Once you have installed and gained access to the `IBM Design kit`, navigate to the `product design` folder. Select the version of Carbon you'd like to use to find the `Carbon Design Kit` file.
-- When updates are available to the kit library a purple `Library Update` badge will appear in the top right corner of your Sketch file.
+1. Follow the instructions on [IBM Digital Design](https://www.ibm.com/standards/web/design-kit/) to get the design kit through Box Drive.
+- Once you have installed and gained access to the `IBM Design Kit`, navigate to the `product design` folder. Select the correct version of Carbon to find the right Carbon Design Kit file.
```
aagonzales: there's not a "correct" version. Some designers might still be using v9 after our V10 release because their development team hasn't upgraded yet.

Maybe something like this:
"Select a version Carbon you are using then the Carbon kit file will be in that folder."

src/content/getting-started/designers/designers.md
```diff
@@ -3,34 +3,32 @@ label: Getting Started
 title: Designers
 ---
 
-<page-intro>The **Carbon Design Kit** is a living, breathing document that contains all of our visual assets (components, iconography, color palettes, grids, etc) aimed to help you work faster, with confidence and thoughtfulness.</page-intro>
+<page-intro>This site, along with the Carbon Design Kit, provides designers with everything they need to get up and running with Carbon.</page-intro>
```
aagonzales: Not sure if I like a direct reference to the site on the site but not that big of a thing.

aagonzales:
Thanks for going in and making this updates Clay!

src/content/getting-started/designers/designers.md
claycrenshaw: revised to say "desired" instead of "correct"

src/content/getting-started/designers/designers.md
claycrenshaw: Changed to "Synchronize" to avoid any possible confusion.

claycrenshaw:
@alisonjoseph Should be good to go.

PR #401, opened 2018-11-14
WIP: update do/dont components

lovemecomputer:
Closes https://github.com/carbon-design-system/carbon-website/issues/354
Closes https://github.com/carbon-design-system/issue-tracking/issues/39

**to do:**
- [x] implement ability to use images with their path written in the markdown file
- [x] fit proper sizing/spacing from @designertyler's feedback
- [x] rename `<Example>` component to `<DoDontExample>` to be more clear
- [x] put in icons as semantic elements
- [x] Documentation: update code example in `CONTENT.md`
- [ ] Documentation: add example for including images as a do/dont component
- [ ] Documentation: update screenshot in `CONTENT.md`
  - @alisonjoseph want to verify with you the proper place to put the new screenshot, since the image url is a `user-images.githubusercontent.com` address, rather than in the project folder
- [x] Content: update proper caption content in `guidelines.md` > `### Do use simple tense` (I put some example caption variations for this preview link)

#### Changelog

- for do/don't example sections, now instead of writing `<page-flex-group>` in the markdown file, the author will use `<do-dont-group>`

**example from content guidance page:**
```
<do-dont-group>
    <example correct=true title="Future Tense" text='"The API returns a promise."'></example>
    <example title='Future Tense' text='"The API will return a promise."'></example>
</do-dont-group>
<do-dont-group>
    <example correct=true title="Past Tense" text='"The limit was exceeded."'></example>
    <example title='Future Tense' text='"The limit has been exceeded."'></example>
</do-dont-group>
```

examples have 2 optional properties:
- `title`
- `description`

both of these will be put in a caption under the white example card. either or both can be omitted.

**Note:**
- these do/don't components currently only support text content, as that is the only usage we have on the site.
- the height of the white example cards is flowing/dynamic, based on the length of text inside of the example card. With different text lengths that lead to a different number of line wraps, they may end up being different heights. if we want to set a particular aspect ratio for the cards, we will need some design + content input.
- do/don't components only support up to 2 items per `<do-dont-group>` as that matches the request + current content on the website.

If either of these 3 final bulletpoints should be changed based on design/content request, we can get that feedback now, or do it in a future Issue/PR! ✨


alisonjoseph:
Can we merge this w/ v10 branch?

lovemecomputer:
@alisonjoseph i initially thought the the checkmark/X would be more presentational, and the CSS implementation matches the Duo implementation… but i think you're right, the icon is the only actual bit of content that would communicate wether it's "correct" or "incorrect" since there is no text label. good idea! i think i'll be able to use @joshblack 's icon component library that he showed me yesterday too!

designertyler:
We will eventually want to have images in the do/don'ts. I don't know of any in the Carbon site today, but they will be coming as we update usage guidelines. 

We got feedback yesterday about text alignment and I got a little more clarification after the demo.

I'm 99% sure it's caused by the gray background that's been added. The text at the bottom of the do/don't and the white card itself should left align with the rest of the text on the page.

Mocked this up to show
screen shot 2018-11-16 at 9 29 44 am

Here's a screenshot from language 
screen shot 2018-11-16 at 9 32 29 am

src/components/Example/Example.js
```diff
@@ -1,18 +1,36 @@
...
   static propTypes = {
     children: PropTypes.node,
     title: PropTypes.string,
+    description: PropTypes.string,
     text: PropTypes.string,
-    correct: PropTypes.bool,
+    correct: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
+  };
+
+  renderCaption = (title, description) => {
```
asudoh: Just an idea; The signature of this method is close to React's function-based components. Does this signal that we can extract this portion as a separate component - If it's used only in `<Example>`, probably a component that is not exported?

alisonjoseph:
@lovemecomputer do you want to go ahead and add support for images in this PR?

lovemecomputer:
pushed latest version i mentioned, looking forward to chatting with @alisonjoseph to decide on the best solution, and to catch up on using her grid code!

lovemecomputer:
thanks @alisonjoseph for helping me with the image support! i'm working on the rest of the to-do items and then we can consider this component done according to the current ask :)

lovemecomputer:
@alisonjoseph i might want to check in with you on monday, i'm getting a weird series of merge conflict issues! @dakahn helped me resolve it, but neither of us knew what we did to fix it, and after changing another file the merge conflict came back!

(this might be because i added the elements/icons package? a lot of the errors are in `yarn.lock`)

alisonjoseph:
Ill fix the merge conflicts again later today or monday, they were small. and after I do just pull the latest, delete your lock file and then re-run yarn.

lovemecomputer:
@alisonjoseph change pushed, thank you for the updates!!!

PR #468, opened 2018-11-26
feat(tabs): add experimental tabs

### LINKED ISSUE #440: Add Experimental Tabs to website
- [ ] Add Experimental component to website
- [ ] Update component status

jnm2377:
Closes #440 

**New**
Adds experimental tabs to website

**Bugs**
~~Styling on website is changing for tabs component.~~
~~1. padding in between the tab lines disappears~~
~~2. in the responsive view, the arrow is inverted (i.e. starts right side up)~~

~~I don't think the `carbon-components` repo has these 2 bugs. Still need to figure out what the problem is. Also should we have less tab examples so that the tabs don't come off the example container?~~


alisonjoseph:
@jnm2377 you might have to add some styles in `component-overrides.scss` to fix the style overrides. We've had to do this with a few components on the website.

PR #492, opened 2018-11-29
Update angular docs

youda97:
Update the installation steps and add a section for our carbon-angular-starter app


alisonjoseph:
thanks! merging in, will be live when we push website next

PR #497, opened 2018-11-30
Typography page

alisonjoseph:
Ref #475

Updates to the typography page and custom components to match new designs

**Missing from this PR**

- tab component
- jump link component

**Need**
- A list of tokens we can drop on the page until we have time to build the fancy tab system. @jeanservaas @joshblack 
- @joshblack can you check the code snippets and font-stacks on the main typography page
- Any content edits from @claycrenshaw (can do after merge if needed)


alisonjoseph:
we should probably just remove the code snippets then right @joshblack ?

PR #597, opened 2018-12-12
V10 back to top

### LINKED ISSUE #566: [Website Component] Back to top button
Need **Back to top** button like below for users to go back up to top of page at any point in the scroll. We can follow DL design exactly.

alisonjoseph:
Closes #566

Adds back to top link to match IDL site. 

@shixiedesign still unsure on icon name for this. Currently using `UpToTop20` let me know if there is different icon I should use instead from the library.


shixiedesign:
IDL uses a custom icon for this. I recreated. Checking with Josh if we can add it to library. Also reaching out to IDL team to add it to the icon library. Waiting to hear back before updating site. 

image

shixiedesign:
Heard back from IDL team. The current icon on the IDL site is out of date. Our back to top (without the vertical line of arrow) is the correct one! 🎉

PR #598, opened 2018-12-13
[RFC] Monorepo setup for website add-on development

joshblack:
Wanted to put up a quick RFC for a monorepo-style setup for the website where we can do add-on development.

Work is definitely a proof-of-concept, wanted to show:

- Setting up the monorepo
- Creating a simple package with JS and styles
- Using package in Gatsby
- Using package in storybook

Work that still would need to be done:

- Add build step to package
- Decide on scss strategy for add-on
- Make sure storybook works
- Fix the Layout where I'm using the package 😂


joshblack:
Done @alisonjoseph! Let me know if I missed anything 😄 Also fixed some issues with Storybook that I noticed when trying to get the example one working.

alisonjoseph:
Thanks @joshblack! I can't seem to get storybook to load? is that working, or something todo still?

PR #610, opened 2018-12-18
chore(devenv): separate task of cleaning cache

asudoh:
This change separates the task of cleaning Gatsby cache from `dev` task. It appears that Gatsby has its internal cache validation mechanism and I guess it works for many cases. Created `clean:cache` task in case it doesn't happen. Wondering if makes sense to the team.

#### Changelog

**New**

- `clean:cache` Yarn task to clean Gatsby cache.

**Changed**

- `dev:external`/`dev:internal` Yarn task no longer clears Gatsby cache.


asudoh:
Yeah I imagined that happened - That's the exact reason why I added `clean:cache`. Thank you for your comments @alisonjoseph!

PR #627, opened 2019-01-02
fix: update Angular components info

cal-smith:
Because we support more things now 🎉


cal-smith:
Sure 👍

lovemecomputer:
awesome, thank you @cal-smith !!

PR #637, opened 2019-01-04
Update icons in component examples (v10)

alisonjoseph:
Closes #https://github.com/IBM/carbon-elements/issues/212

Pulls in latest html from carbon-components with the icon updates. 

FYI: Once v10 is live we can remove the html from this repo, if there is no html file the page will pull directly from carbon-components.


shixiedesign:
checked the X icon looks good!

Issue #662, opened 2019-01-09
Carbon X Website Component: Content Switcher


jeanservaas:
A content switcher (sub-tab) component is needed to switch between background color themes in several places on the usage tab on the Color Page for v10 launch:

> The tab component is not appropriate for this interaction and anchor links are not ideal

> This component needs to be sticky as it affects large areas of content beneath the banner tab

> We need to surface the four color themes for our Color guidance to make sense

<br>

## Design Specs
https://github.com/carbon-design-system/carbon-website/issues/638

jeanservaas:
Approved!

PR #670, opened 2019-01-10
feat(color-token-table): added custom component color token table + content switcher

### LINKED ISSUE #618: Website_Component_Color_Token_Display
## Color Token Display Component for Color Elements Usage page

Sketch file: https://ibm.box.com/s/chioa7y5gzz91i7iaf8oepryw67vgggu

<b>Purpose: </b>

> This component marries up swatch copy functionality of current Carbon with a table of color tokens; it will live under the Usage tab on the Color Elements page. 

> Tokens values change depending on which color theme the user chooses with the Content Switcher component above (spec'd out in issue # XXXX )

> This component is shown at each breakpoint and also uses the right aligned Overflow experimental component 

<b>Specs: </b>

specs

<b>Breakpoints: </b>

1584

color_token_component 1584

1056

color_token_component 1056

672

color_token_component 672

Mobile

color_token_mobile_revised

<br>

<b>Usage tab in context: </b>

1312

color in ui_swatches_1312

marijohannessen:
Closes #618 

Design is ready for review, but just using "dummy" data right now as I wait for the new themes-package to be released. Will update this PR when that happens. 

I couldn't get OverflowMenu to work properly without overriding some styles, is this a known problem with the experimental OverflowMenu @alisonjoseph ? 

This PR also includes the Content Switcher from this issue #662 

Ready for review here: https://deploy-preview-670--carbon-website.netlify.com/guidelines/color/usage


src/components/ColorTokenTable/ColorTokenTable.js
```diff
@@ -0,0 +1,142 @@
...
+            <Switch name="gray-10" text="Gray 10" />
+            <Switch name="gray-100" text="Gray 100" />
+            <Switch name="gray-90" text="Gray 90" />
+          </ContentSwitcher>
+        </div>
+        <div className="ibm--col-lg-7 ibm--offset-lg-4">
+          <h3 className="page-h3">Core color tokens</h3>
+        </div>
+        <div className="ibm--col-lg-12 ibm--offset-lg-4 ibm--col-bleed">
+          <table className="page-table">
```
marijohannessen: Planning on componentizing these tables once I see what the format will be on the data from `@carbon/themes`

src/components/ColorTokenTable/color-token-table.scss
```diff
@@ -0,0 +1,128 @@
...
+}
+
+.color-token-value .bx--overflow-menu {
+  margin-left: $spacing-xs;
+}
+
+.color-token-value .bx--overflow-menu-options {
+  left: -9.25rem;
+  top: $spacing-xl;
+}
```
marijohannessen: If I don't add the CSS above for the `bx--overflow-menu-options` the menu floats to the right by 100px

src/components/ColorTokenTable/color-token-table.scss
```diff
@@ -0,0 +1,128 @@
...
+
+  @media all and (max-width: 580px) {
+    width: rem(48px);
+    height: rem(48px);
+  }
+}
+
+.color-token-table .page-table {
+  background: $ibm-colors__white;
+  max-width: rem(962px);
```
marijohannessen: Not sure if we need this value hard-coded, taken from the Sketch file

jeanservaas:
## Color token component comments (looking so good!!!)

#### Width

- we talked about this in person and I think you fixed it, width should mirror width and breakpoint behavior of the type spec component built by @alisonjoseph 

(I think it's 960px / 12 columns / with 16px hanging gutter on both sides @ 1312 breakpoint & 1150px / 12 cols / hanging gutter @ 1584 breakpoint) 


#### Dividers 

- rules are 1px #dcdcdc (Gray 20) ... looks like they're 2px right now just showing Gray 10 background)

image

#### Alignment

- please left-align category title with copy in the table (token looks good but Role and Value look jogged)

image


#### Content

- this is a super minor content thing, but that divider dash in the value column is actually an em-dash; this may be a markdown thing

—  (shift + option + dash)

- same thing... this might just be us in markdown; just don't want the last item in the usage list to have a ; after it

image


#### Content switcher

-  we got some feedback from Mike on this guy and for your ref. below is the spec'd issue; we were really not sure if we could get anything other than anchor links in so I was REALLY excited to see anything here at all!!!!!! 

https://github.com/carbon-design-system/carbon-website/issues/662

src/components/ColorTokenTable/color-token-table.scss
alisonjoseph: I just took a look at the overflow menu page and the styling is off there also. @asudoh any ideas on this? 
screen shot 2019-01-10 at 2 21 44 pm

jeanservaas:
## Color component comments:

Looks awesome, small thing:

- when the component goes full screen in mobile, i was wondering if the tiles could assume the size of the tallest one (i.e. since the first tile has four lines in usage)... if this is too hard to do, etc... just lmk — I just think it would look tidier

<br>

preview link
image

<br>

spec
image

## Content switcher:

- not sure if you can pick up the content switcher issue #662 in git next or if this is going to stay the content switcher; in the meantime, the hover color for it should be the $hover-ui token for the white theme (#e5e5e5)


image

image

alisonjoseph:
This is looking so good! 🎉 Would be good to add a line about this component to `docs/CONTENT.md` , there should be a section already on page specific components.

src/components/ColorTokenTable/color-token-table.scss
```diff
@@ -0,0 +1,163 @@
...
+  // }
+}
+
+.color-token-table__theme-switcher--sticky {
+  position: fixed;
+  top: 64px;
+  z-index: 9999;
+  max-width: 1164px;
+  box-shadow: 0px 2px 6px 0px rgba(#000000, 0.2);
+
```
marijohannessen: This part with the media queries is pretty gross, trying to think of a better way to do this

jeanservaas:
Content Switcher looks great! Tiny things:

- wondering if we can add a little easing to the drop, when the component becomes sticky; it drops a little hard and it looks like it makes the page header jump slightly in Chrome. (@shixiedesign is saying something about 20px displacement) I'm going to send you a little motion test shixie did, it's an outdated design but the motion easing / specs would still apply)

- wondering if we could extend hover state by a couple of pixels so it actually covers the rule... _this is so tiny_ but there's something weird about how the hover state doesn't cover the rule but the rule disappears anyway on hover.

image

Issue #629, opened 2019-01-02
Carbon X Release: Component — Website Side Nav 


jeanservaas:
## Update Website Side Nav universally
We need to create a variation of the Product UI Shell for digital experiences; our goal is to align style and interactions and geometry across Carbon, the IDL and the Digital sites.

### Design Issue
https://github.com/IBM/carbon-components/issues/1559

See specs for visuals:

 - [x] Use Gray 100 theme tokens for homepage side nav; White theme tokens for the rest of the site

- [x] All side nav typography should be tokenized 

 - [x] Update spacing: all link item heights will be 32px to save space (not 48px)

 - [x] Add indents for nested items 

 - [x] Remove tinted background behind nested items (only selected category will get selected background token

 - [x] IBM Product Design (on the header bar) should be 1 rem in size (not .875 rem)

PR #681, opened 2019-01-11
fix(side-nav): style updates to side nav

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/629

Makes changes to side nav heights, spacing, colors and typography.

Also, addresses hover issues, as well as restyling the footer section.

Light theme has been implemented as well


alisonjoseph:
Quick question, did I hear a rumor that the focus states are getting updated to 2px instead of 4px?

jeanservaas:
Looks amazin'

we talked about this, just wondering (in Chrome) if the background color could change to a G90 (282828) ONLY on the homepage since it's dark... IDL looks like they do this on the homepage for when the scroll jogs the container.

image

otherwise DONE! boom.

Issue #683, opened 2019-01-11
Image Component


alisonjoseph:
### Image component requirements (revisited by Shixie)

- [ ] Caption support. Ref Jeanie's comment below for spec https://github.com/carbon-design-system/carbon-website/issues/683#issuecomment-454586014
- [ ] Need a way to specify how many columns an image should span. 
We only need 4 options: `lg=12 col` ,`md=8 col`, `sm=6 col`, `xs=4 col` (these are values at max breakpoint. More detail see below image or https://github.com/carbon-design-system/carbon-website/issues/524)
![image](https://user-images.githubusercontent.com/15144993/51349875-57013d80-1a6c-11e9-97c4-7ca17d224fcb.png)

- [ ] An option to turn on-off "on grid". fixed (has a max pixel width and max-width 100%) vs. fluid (scales on grid)

- [ ] Default images in Markdown eg. `![caption](image-url)` should have `md` size if landscape, `sm` size if portrait.

**post-THINK**

- [ ] A way to turn on the ability to "onClick to view large" in a modal
We would need for component spec images coz specs are usually small and hard to read, not necessary for all images). The ideal experience is to expand the image in a modal (instead of in a new Tab because it will be missing background). This is also good in preparation of future iteration of "live spec"

jeanservaas:
Hey @alisonjoseph 

I see the caption support box checked above and I was wondering if you added a caption piece to the images... if so I want to review it here vs. making a new issue for it.

This is your caption update, just a couple of feedbacks:

- token we should be using is $body-short-01 (so it's the 1px at the 18px line height; current line height looks bigger)

- also, we need 8px of padding above the caption (or below the image); right now it is too tight on the image

- Caption color looks a little light right now, it will be #171717 (G100) just like body copy

- and we need to control the line width with some right padding (see example) (line width will match line width of Do / Don't component

![image-4](https://user-images.githubusercontent.com/43144260/51215650-3e1b4f80-18e7-11e9-80bb-f22d8f775d97.png)

alisonjoseph:
@jeanservaas will push up a fix now! thanks for the feedback

shixiedesign:
Reviving this issue because the Image Component turns out to be a big one. Here are the updated list of features needed for <image></image> component: 
- [ ] caption styling
- [ ] sizing support (at max breakpoint, our 4 widths options are `lg=12 col`, `md=8 col`, `sm=6 col`,` xs=4 col`)

[...]

PR #713, opened 2019-01-15
fix: image captions

alisonjoseph:
cc @jeanservaas 
https://github.com/carbon-design-system/carbon-website/issues/683

updates the image caption styling

FYI markdown to add a caption to an image, followed by a line break, followed by italic text.

```
![Structure and spacing measurements for a Definition Tooltip](images/tooltip-style-3.png)

_Structure and spacing measurements for a Definition Tooltip | px / rem_
```

### DIFF
--- a/src/styles/_page.scss
+++ b/src/styles/_page.scss
@@ -115,6 +115,7 @@ button,
 //images 
 .gatsby-resp-image-wrapper {
   background: $ui-02;
+  margin-left: 0!important;
 
   @include breakpoint('lg') {
     margin-left: calc(25% + .5rem)!important;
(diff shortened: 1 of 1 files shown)


shixiedesign:
@alisonjoseph Image captions are updated on some places on this page but not here. Looks like the markdown syntax doesn't work consistently. Am I doing something wrong here? 
image

alisonjoseph:
I'll take a look, svgs are rendered differently from other images within Gatsby so that is most likely the issue here. Your markdown is correct.

shixiedesign:
Yep PNG (bottom version) is working but SVG is not!  
image

alisonjoseph:
Just go ahead and add the captions where you need them and I'll submit a fix later today for svg captions.

shixiedesign:
FYI image caption would also need to work multi-lined. Current markdown of below caption is: 
> _1. Header_
_2. Global sidenav_
_3. Local sidenav_

image

Issue #710, opened 2019-01-15
Certain components should be full screen at mobile (sm) breakpoint 


jeanservaas:
## No margin at mobile breakpoint for these components:

- @alisonjoseph not sure if it's just better to deal with each of these situations in an issue related to that specific component vs. a list like this... if so, you can close this issue we can do it that way (since we do already have issues for a few of these components).

- Type spec component, overview display component, color token display component and code snippet (on certain pages) are all doing this correctly; these are the components that are not:

<br>

- [x] Banner tabs: no margin at mobile (already dealt with in issue #708)

<br>

- [ ] Icon display: no margin at mobile

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/c913535f-e9fe-4228-9695-4a335ce25a81)

<br>

- [x] Component display window / code snippet

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/3046371e-7d11-455d-9060-d3866de6f2f5)

<br>

- [x] Table: component: no margin at mobile

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/d18ef4b1-aa56-457b-8b5d-d0ee2596b75f)

alisonjoseph:
Icon page hasn't been done yet, will leave that for that issue

Issue #708, opened 2019-01-15
Website Banners not to Design Spec


jeanservaas:
## Website Banners are not to spec

> There are several issues with the banner (some design, some display) that are happening across browsers, which is why it's getting a higher priority tag

<br>

## Banner update hit list

- [x]  Align tabs correctly: text within tabs need to align to the text on the rest of the site and the tabs need to hang into the gutter

<br>

<b>Spec:</b>

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/b312839c-4baf-4e97-92f3-23854a6b5f4a)

<br>

- [ ]  Mobile banner tabs should follow specs in this issue: https://github.com/carbon-design-system/carbon-website/issues/578#issuecomment-450960617

- [x]  At tablet (Md) breakpoint, the intro sentence at the top of the banners in the Guidelines section are spanning the entire browser); We need right margin padding here like every other breakpoint

<br>

<b>Problem:</b>

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/c350c0d1-0bdc-488f-8585-f29b1299b589)

<br>

<b>Spec @ 672:</b>

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/32b2b357-8c04-4527-8078-e4d69c2f517d)

- [ ]  The banner H1 type should be responsive (use fluid token: $Display-01)

- [x]  Add hover states to unselected banner items (they're not working now in any browser)

<br>

<b>Problem:</b>

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/09a4d2bb-414a-4be9-8f76-4c155de0d042)

<b>Spec:</b>

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/dc4d3c31-d885-4b57-a76c-1e3debb7d8d6)

<br>

- [x]  Fix browser display issue with the banner H1: right now the H1 looks like it's pulling in Plex Extra Light in Firefox and Chrome... it looks right in Safari... I'm wondering if the token change will fix this issue?

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/c37072b2-a016-4a60-a39c-1f57fe3daff9)

- [ ]  Ultimately tabs should be sticky to the top of the browser on scroll (as soon as you begin to scroll up again, top bar appears; This may not be doable until post-THINK

[...]

Issue #578, opened 2018-12-10
Enhancement [sticky tabs]: Top nav and tabs stacking behavior 


shixiedesign:
Mike provided below assets describing Tab's stacking behavior. Digital's site is consistent with this, and we should aim to build it in, if not this release, potentially shortly afterwards.

![nav-tab-stacking-full.gif](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/16f4472c-38b2-4f5b-9c83-f31a5e39bf8d)

See full res videos of above: 
https://ibm.box.com/s/np46yjxa0xa5je1ifqkbh2bigxfhevjo
Sketchfile: https://ibm.box.com/s/yshar9cm6tnbr8zdzjrwvstihf0anhjq

alisonjoseph:
How will this work with more than two tabs? I think this page is the worst on mobile currently https://www.carbondesignsystem.com/getting-started/developers/vanilla

alisonjoseph:
@jeanservaas makes sense, thanks! 

_They don't do that on the site currently because they aren't using the tab component, they aren't technically tabs, they are links to new pages._ 😉

jeanservaas:
Sorry @alisonjoseph -- i realized that I massively needed to provide you a visual on that one. It's an outlier because the banner tabs don't really function like the experimental tab component in that it spans full screen... so it got a little weird at the tablet breakpoint and @shixiedesign and I had to look at it.

sketch file: https://ibm.box.com/s/tmtrrtphiio2g7duvb4uwchbvjd6h8z6

![image](https://user-images.githubusercontent.com/43144260/50614913-10c2b080-0ea8-11e9-8ce5-60803757cafb.png)

mjabbink:
@jeanservaas and @shixiedesign are we going to stick with smaller ones (like current next.carbon) and go upt to three tabs before going to the dropdown. I think it looks good being smaller versus the 50/50. What do you all think?

jeanservaas:
@lovemecomputer 

After mike saw the stacking spec, he wanted a slight pivot in the design... he did not want us to go to 50% of the screen with two tabs at mobile anymore.

- [ ] Two banner tabs will not go 50/50 on the screen at mobile anymore but will stay as they are now on the carbon site

- [ ] Our tab component (the in-page component seen on the motion page) the uses a minimum width to make tabs all equal width if they are under a certain character count; it would be nice to do something like this with our banner tabs as well (or add some light JS)

[...]

PR #714, opened 2019-01-16
Responsive fixes

alisonjoseph:
closes https://github.com/carbon-design-system/carbon-website/issues/710
ref https://github.com/carbon-design-system/carbon-website/issues/708


jeanservaas:
Thanks for making a new issue to track this:

Banner tabs look much better but we should still be going to dropdown per this issue after two tabs so stuff like this doesn't happen:

image

Here is the design spec for that:
https://github.com/carbon-design-system/carbon-website/issues/578#issuecomment-450960617

alisonjoseph:
Yes there is a separate issue for that, will be handled next sprint. https://github.com/carbon-design-system/carbon-website/issues/578

PR #716, opened 2019-01-16
V10 app switcher

### LINKED ISSUE #510: [Dev]: App Switcher
@claycrenshaw @designertyler @jeanservaas @shixiedesign 

We need specs and content.

Or we can just copy https://w3.ibm.com/design/language/ ?

aledavila:
Closes #510 

Adds website switcher panel

#### Changelog

**New**

- website switcher
- only 1 panel open at a time

**Changed**

- toggle on click function to be reusable for both panels


packages/addons-website/scss/components/_website-switcher.scss
```diff
@@ -0,0 +1,61 @@
...
+  z-index: 1000;
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 0.2);
+  will-change: width;
+  transition: width 0.11s cubic-bezier(0.2, 0, 1, 0.9);
+  overflow-y: auto;
+  overflow-x: hidden;
+}
+
+.#{$prefix}--website-switcher--expanded {
+  width: mini-units(32);
```
tw15egan: Can we change the animation to use `transform` rather than width? Transform makes the browser use the GPU, which should give a slightly smoother animation. 

```css 
.#{$prefix}--website-switcher { 
   transform: translate(16rem);
}

.#{$prefix}--website-switcher--expanded {
    transformL translate(0);
}
```

alisonjoseph: Are we still using mini-units? 🤔  I don't think its used anywhere else on the site, can we use rem(px) instead.

alisonjoseph:
Storybook doesn't appear to be showing the switcher? 

https://deploy-preview-716--carbon-website.netlify.com/storybook/?selectedKind=Website%20Switcher&selectedStory=Website%20Switcher&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

aledavila:
Thats weird I copied exactly what you had for the footer. Do I need to build it maybe. How can I debug storybook locally ?

alisonjoseph:
@aledavila you should be able to just run `yarn storybook`

shixiedesign:
First link **IBM Design System** is not clickable. It should point to `www.ibm.com/design/language`

image

aledavila:
@alisonjoseph Thanks I'll see if I can solve it if not I'll ping you

jeanservaas:
Hi @aledavila !

Type size is too large right now for nav items also, it should be pulling $heading-01 (right now it looks like it might even be $heading-03 or something) — so it's semi-bold (600), 14px type

Colors and spacing look right to me -- just missing that hover state on the top item 

image

packages/addons-website/src/components/WebsiteSwitcher/WebsiteSwitcher-story.js
```diff
@@ -0,0 +1,21 @@
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import WebsiteSwitcher from '.';
+
+storiesOf('Website Switcher', module).add(
+  'Website Switcher',
+  () => (
+    <WebsiteSwitcher
```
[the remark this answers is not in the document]
aledavila: Ok done. I see the issue I guess it's because it's technically hidden until you click the switcher in the header. Thanks

jeanservaas:
Looks great!

- One super small thing, and I may be wrong, just want to make sure the rule is equal-distant from both links, (8px spacer); in a couple of browsers it just looks like the lower link item might be like a pixel further away.

image

- Also, focus state is 2px not 4px -- tj had the same problem in the nav

other than that boom, all good from me

aledavila:
No worries let me fix 😆

aledavila:
Ok got everything @alisonjoseph @jeanservaas 

You were right @jeanservaas it was 1px off since I hadn't accounted for the height of the line. But fixed that now

alisonjoseph:
Looks good to me! I think the focus state still needs to be updated to 2px though? Also would be nice to drop the switcher into the header stories if possible.

aledavila:
Yea I'll put it in the story. But the focus state is 2px on the panel itself. The focus on the button is coming from shell in components so would need to change that there and I'm guessing if thats the case most focus states in shell need to change ?

aledavila:
Thanks good to know wasn't sure if we overrode or changed it there. Thanks

jeanservaas:
Looks good to me!

PR #753, opened 2019-01-18
fix(table): make sure table does not go full width of browser

tw15egan:
Fixes tables on large screens


tw15egan:
@alisonjoseph Alright, moved it a bit. Yeah, the tables are a bit weird. I needed to keep the `min-width` even though it is larger than the `max-width`, otherwise the tables would render too small (or too large like before)

shixiedesign:
Table length looks great! I'm noticing the body copies and table don't **start** on grid though...is this an overall grid issue? 
image

alisonjoseph:
@shixiedesign the site is using the new grid code, so if everything is off then it would be an issue upstream in the elements repo. Where is that overlay you are using coming from?

Issue #694, opened 2019-01-14
Create intro copy for Components Overview page


claycrenshaw:
We don't currently offer users any introduction to the concept of "components" in our system. The nav item "Components : Overview" hints at such content, but is actually just a visual index of the components in the system.

Proposed: Add some intro copy to the top of that page to introduce the concept of components, along with some top-level info about how they fit into Carbon.

claycrenshaw:
First draft for feedback: 

Components can be considered the “base unit” of the design system. Each component is designed and coded to solve a specific UI problem, such as presenting a list of options, enabling submission of a form, providing feedback to the user, and so on. All of the components in Carbon have been designed to work harmoniously together, as parts of a greater whole. 

Component design files are delivered via the design kit, and the code for components is available in the repo for the respective javascript framework.

aagonzales:
**Feedback**
- "Components can be considered the “base unit” of the design system." I think we should project more confidence with this statement. There's no uncertainty about it. "Components **are** the base unit of the design system".
- I think you can kill the last paragraph entirely. Those topics are addressed in getting started and people are most likely viewing this page to view the assets on the website not in those different contexts. 
- I would check with dev on the term "javascript framework" I think it would be just "framework"
- I think "Component index" would also work for the title, keep it simple.

chrisconnors-ibm:
**Design System = Design Language, Tooling (Design Templates _and_ code components), and Activation and Outreach** - these are three co-equal things

claycrenshaw:
"Components are one of the key building blocks of the design system."

PR #755, opened 2019-01-18
Component overview edits

claycrenshaw:
Minor edits. 

New PR without Patterns Overview page attached.


chrisconnors-ibm:
has this:
`Component design files are delivered via the design kit, and the code for components is available in the repo for the respective javascript framework.`
moved somewhere else? that seems like a worthwhile definition of what we mean.

claycrenshaw:
@chrisconnors-ibm That paragraph was removed per feedback from @aagonzales. . . https://github.com/carbon-design-system/carbon-website/issues/694

jeanservaas:
Sorry this is late to the party, but I think we could just say "Visual Index" instead of "Visual Index of Components."

PR #764, opened 2019-01-21
Editorial pass - Accessibility

claycrenshaw:
Minor edits for style, clarity, flow, grammar. Added index anchor links.


jeanservaas:
**Overview tab:**

- Avoid changing weights (i.e. using bolds) within the paragraph as we're not doing this treatment in other places; also, it's a forced bold, it's not pulling Plex semi-bold

image

- use H4 class for the categories under "Blind Users," "Low-vision users," "Color-blind users," "Deaf and hard of hearing users," "Physical disabilities," "Cognitive disabilities" (it will add some visual texture).

**Color tab:**

- Don't bold this:

image


- We don't use 19px semi-bold type in our system

image


- Use H4 class for these because they are so short (Shixie and I have been starting to do this when paragraphs are very short and a lot of H3s are close together, the spacing gets awkward)

image


Instead of this image, use the attached:

image

[color_blindness_new.png.zip](https://github.com/carbon-design-system/carbon-website/files/2814803/color_blindness_new.png.zip)


- Could you make "Low-vision disabilities" an H4 (above the image) and make the source (Source: NoCoffee - Visual impairments) the caption

- Please use this image instead:

[Low_vision.png.zip](https://github.com/carbon-design-system/carbon-website/files/2815003/Low_vision.png.zip)

@aagonzales talked to @designertyler about going in and making these all tiles

image

**Keyboard Tab:**

- these links do not need underlines because they are not in a paragraph... may need to check with alison to see if we have a style for this:

image

**Developers Tab:**

- unbold this type

image

alisonjoseph:
are these changes addressed here? https://github.com/carbon-design-system/carbon-website/pull/809 if so go ahead and close, if not ignore me

Issue #508, opened 2018-12-03
[Dev]: Motion Page


alisonjoseph:
Page design: https://github.com/carbon-design-system/carbon-website/issues/476
Components: https://github.com/carbon-design-system/carbon-website/issues/473
Reviewed content (optional): https://github.com/carbon-design-system/issue-tracking/issues/78
Vimeo links of video: https://github.com/carbon-design-system/carbon-website/issues/657

Sketchfile: https://ibm.box.com/s/0yruurkwto8zo6v9tgl86c2l7t9r2ir7
Video & image assets: https://ibm.box.com/s/wyy9df51y4xwljd7xnkl3am3759h49rj

### Quick notes: 
1. Video assets have do/don't designs baked in. This is so that only 1 video component is needed for website, and motion videos should not exist outside website without the do/don't coloring. 
2. Removed checklist component. Only using list. That's one less component to develop. 
3. Clay is waiting to review content after Tabs component is build.

shixiedesign:
Note: Content on current design files (#476) are the same content live on IDL site. We can go ahead a build the page with it, and when Clay comes back with reviewed content, we can update.

alisonjoseph:
@lovemecomputer is this something you can tackle this sprint since you are already working on the tab component that's needed for this page?

claycrenshaw:
Please loop me in on this as MD is created so I can do an editorial pass before we publish. Thanks! @lovemecomputer @shixiedesign

alisonjoseph:
@shixiedesign @jeanservaas need vimeo links for this page for @lovemecomputer

shixiedesign:
@alisonjoseph @lovemecomputer all the vimeo links are here: #657

lovemecomputer:
thanks shixie!

PR #773, opened 2019-01-23
[WIP] Motion content

lovemecomputer:
ref https://github.com/carbon-design-system/carbon-website/issues/508


lovemecomputer:
ooop my local repo wasn't building after merging in the latest changes from v10, so i went branched off from a previous commit to get this PR for content. (either that or it's bc v10 doesn't incorporate those commits from the other branch?) will verify or resolve this all before merge!

alisonjoseph:
.example-content needs a `width: 100%;` to fix the hidden images.

the images inside the tabs are getting cut off, did you pull these from the box folder? Sync up with @claycrenshaw and @shixiedesign on what images go inside the tabs, I think they should be in the box folder in the issue.

lovemecomputer:
@alisonjoseph i have a separate PR up addressing the CSS fixes here: https://github.com/carbon-design-system/carbon-website/pull/779 , but i can just include those fixes into this PR if you prefer! it would def make this preview easier

also yeah, the images i'm getting from the Box folder are what you see here, and there aren't many assets available. i'm prepping all the markup to hand off to @claycrenshaw / @shixiedesign so they can plug in final assets!

shixiedesign:
Thanks @lovemecomputer!! So glad to see this coming together. I can certainly help with putting in the final assets. It's probably gonna be easier since I know the content & videos well. Let me know when I should start doing it. Looks like you still have WIP tag on this so I am waiting.

Issue #770, opened 2019-01-22
The banner H1 type should be responsive


alisonjoseph:
The banner H1 type should be responsive (use fluid token: $Display-01)

PR #777, opened 2019-01-23
fix(banner): ensure banner text is responsive

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/770

Ensures `Banner` headings are responsive. The type classes were already added, so it simply needed the hardcoded values removed


jeanservaas:
This responds well, and this is the banner behavior the digital (erica's team) is using... just a quick question... what expressive token is being used in the banner. At large and tablet breakpoints our type size is so much smaller than IDL, see below:

image

alisonjoseph:
https://github.com/carbon-design-system/carbon-website/issues/770 $display-01

jeanservaas:
yup, talked to erica -- i'm good with this, the token is right, if I have a size issue i can change it in the .json file

Issue #727, opened 2019-01-16
Secondary anchor link pattern hover not working (not to design spec)


jeanservaas:
## Secondary anchor link fixes

> Right now this pattern is only on the Layout page, but we will use it more in the future

> This is both a bug and a design fix

> The line height is too big: the type should be pulling $body-long-02 (16px type over 24px line height)

> when you trigger the hover, but your cursor is not over the type, the anchor link will not work (even though the type has a hover color); This is happening in Safari, Firefox and Chrome

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/d6459906-94ee-402b-a261-5f0d8644e994)

PR #778, opened 2019-01-23
fix(page-links): adjust padding and margin so entire link is clickable

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/727

Adjusts the page links on layout page so that the entire link is clickable


jeanservaas:
Couple comments:

**There are actually two different sizes for the anchor link component (the one on the top of the page is $heading-03) It's size is correct, but it was having the link issue which it looks like you fixed.**


- [ ]  just change first anchor link component size back to what it was, $heading-03 (the fix on the link state looks good)

This is the 1st anchor link component (all the ones at the top of the page)
image

**The secondary anchor link component is only on the Layout page (that's why it's confusing), but we will use it more in future; the link fix looks good but the line height is still too much**

- [ ] line height on secondary anchor link is $body-long-02 (16px (1rem) / 24px (1.5rem) 

all the rest of these (below the top set of anchor links on the layout page only) are secondary anchor link component.

image

tw15egan:
Updated 👍

PR #784, opened 2019-01-24
fix(spacing): Adjusted spacing across the website

### LINKED ISSUE #739: Website spacing adjustment
Need to achieve consistent spacing through out the site. Here are the rules for when to have a margin and when not to. Please note that **margins are only applied to type elements. Components such as image, cards, table, component example window, and code snippets gets 0 margins all around.**

# Default Heading Margins

### H2 
Margin-top: 80px
Margin-bottom: 24px

### H3 
Margin-top: 48px
Margin-bottom: 16px

### H4 
Margin-top: 48
Margin-bottom: 0px

### Paragraphs 
Margin-bottom: 24px

# Rules for Situational Changes

1. Heading (H2, H3, H4) always have a margin-top (ref size below), so any margin-bottoms of the previous thing should be removed when a Heading follows. 

2. When a Heading is alone (not followed by a paragraph), it gets a 24px margin-bottom. When a paragraph is alone (missing a heading), it gets 48px margin top. 

3. If a component is followed by a component, there should be 24px in-between. (prefer to think of this as adding a spacer, and not changing component margins.)

### Example

Below example demonstrates two of above rules. Because there's a H3 following the H2, H2 in this situ looses its margin-bottom, and we only see H3's margin-top. The second H4 (Public Methods) is followed by a table component, so 24px margin-bottom is added to H4. 

image

### Example 

Example for rule no. 3. Spacing between two images is 24px. Caption is treated as part of the image, so gets 0 margin-bottom.

image

### Example

image

### Index anchor links
Margin-top: 80px
Margin-bottom: 80px

### Index anchor links secondary (should be treated as a paragraph basically)
Margin-bottom 24px

Boxnote of this content for reference: https://ibm.box.com/s/xqs7yrs2fqeh8w4c5mj1e5v1talipcqw

marijohannessen:
Closes #739 

Did my best attempt to adjust spacing across the website. Unfortunately, this is easier said than done because of the grid and therefore not being able to target certain elements/patterns (for example when an image comes after an H2 etc). 

**A couple of things:**

- I did not do any spacing adjustments on images and captions as this will be addressed with the new image component.
- Was unable to adjust the spacing on the iframes on the Layout page as they are using padding to be displayed, so unsure how to approach this
- I guessed a lot for the spacing on the Glossary page, would love to have some input here @shixiedesign @jeanservaas 
- There are some hacky CSS in here now because of certain outliers, are we okay with this @alisonjoseph ? 

Also, I did one page at a time and I might have fixed something on one page that affected another page, so will need to double check all the pages to make sure it looks okay.

### DIFF
--- a/src/components/DoDontExample/do-dont-example.scss
+++ b/src/components/DoDontExample/do-dont-example.scss
@@ -9,8 +9,6 @@
 
 // Example
 .example {
-  margin-bottom: rem(36px);
-
   @include breakpoint('sm') {
     width: 100%;
   }
@@ -128,6 +126,7 @@ p.example__description {
   @include font-smoothing;
   margin: 0;
   padding: 0;
+  padding-bottom: rem(24px);
 }
 
 .example.example--full-width {
--- a/src/components/DoDontExample/do-dont-example.scss
+++ b/src/components/DoDontExample/do-dont-example.scss
@@ -9,6 +9,8 @@
 
 // Example
 .example {
+  margin-bottom: rem(24px);
+
   @include breakpoint('sm') {
     width: 100%;
   }
@@ -118,7 +120,7 @@ p.example__text {
 .example__title {
   @include type-style('body-short-01');
   margin: 0;
-  padding: 0;
+  padding: 0 !important;
 }
 
 p.example__description {
--- a/src/components/DoDontExample/do-dont-example.scss
+++ b/src/components/DoDontExample/do-dont-example.scss
@@ -117,10 +117,10 @@ p.example__text {
   padding: 8px 0 0 0;
 }
 
-.example__title {
+p.example__title {
   @include type-style('body-short-01');
   margin: 0;
-  padding: 0 !important;
+  padding: 0;
 }
 
 p.example__description {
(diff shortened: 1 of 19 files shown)


src/components/DoDontExample/do-dont-example.scss
```diff
@@ -119,12 +119,13 @@ p.example__text {
 .example__title {
   @include type-style('body-short-01');
   margin: 0;
-  padding: 0;
+  padding: 0 !important;
```
alisonjoseph: I think if we just change this to `p.example__title` we won't need the important?

shixiedesign:
Noticing these issues: 
- [ ] Video (i guess this is the iframe issue) need a fix though. @alisonjoseph any suggestions? 
image
- [ ] List items line height should be 1.5rem, not 1.75rem. List items also have a tiny indent. Let's remove it if possible. 
image

shixiedesign:
- [ ] List should not have margin-top and margin-bottom. Only getting margin-bottom from the thing above it. Below comes from **Guidelines>Accessibility Color tab.**
image

- [ ] Under **Getting started>Designers** this looks like there's too much spacing. Scroll down for spec'ed version:
image

Ideally want above to look like this. Under first level bullets, there should be 8px gap. Second level bullets should be indented 16px from type. Images are indented but that's a content issue we can go in to fix.

image

- [ ] In *Getting started > Developers*. When paragraphs are sandwiched between components (code snippet in this situ), there should be a 24px spacer above the paragraph.

PR #799, opened 2019-01-26
Fix(icon, motion, type): elements images update

shixiedesign:
Updated images in: 

* Iconography/Usage
**Corrected image size** See below for current: 
image

- Motion
**Curves images. Small issue here - images are not laying out as 2 col.**

- Typography
**Fix currently broken svg** See below for current:
image


alisonjoseph:
Looks good, there are a couple broken/missing images on the motion page still, wasn't sure if that was still a WIP?

shixiedesign:
@alisonjoseph motion page content WIP! Working on it now and should be fixed in the next PR.

PR #801, opened 2019-01-27
Motion page content

shixiedesign:
All motion content is here! However, the page suffers from a few major issues:  
- [x] videos do not work in tabs
- [x] videos do not work in `<grid-wrapper>` 
- [x] video caption styling is wrong
- [x] ~~spacing around videos are too big~~ Resolved with do-don't component
- [x] how to use <grid-wrapper> to show content as 2 columns?


shixiedesign:
Oops looks like we have tab in a tab! @alisonjoseph I'm gonna hold off on changing anything until hear back from @lovemecomputer . 
image

lovemecomputer:
@shixiedesign @alisonjoseph looking at this, it seems like the biggest issue is just that it's fragile to get markdown working within custom component markup. unfortunately, at least at the moment we mainly have to avoid indenting anything—the Markdown process tries to process some of the indentations and that will break the components.

i'll test a few more things to see what else can be fixed up before handing this back to you @shixiedesign !

shixiedesign:
We are almost all done! The only issues I think we should try to fix (these might not be your issues) are 

- [ ] the video's huge margin top and bottom 
- [ ] video captions styling (should be same as image caption style)
image

Question for @lovemecomputer is it easy to give an option to hide the do-don't component's icon and color bar? If not, I can go re-render these videos without the icons & color bar (see below)

image

shixiedesign:
Hi @lovemecomputer I think if we can fix these 2 issues, motion page will be done. Let me know if these things will be an issue. Thanks!  🙏 

- [ ] Video caption is much better but needs to be even shorter. IDL actually use the same caption width as only half of the do-don't example. Basically the width of a caption is always the same no matter how big the video/image above it is.
image

- [ ] Video cropped inside tabs. Looks like video is either not aligned right or too big, so left side of the video is missing a little bit, maybe 2-4 pixels. 
image

Issue #724, opened 2019-01-16
Features/Article component feature enhancement


jeanservaas:
## Features/Article component: spec update

> This is more of a feature change at this point (would be great to get it done before Think because it is so prominent on the homepage)

<b>Issue:</b> 
Mike asked for more separation between the article images for clearer legibility and a better experience and I think he's right (my bad); it's not a great user experience in terms of clarity right now, especially in tablet and mobile.

Also, the current article component doesn't seem to be aligned to the grid right now (i.e. it's not the same width as the "Start Designing/Dev'ing" component, it's like 10px shorter), and this design update will fix that. 

- [x] Add 16px gutter between featured article items (component hangs into gutter 16px only on the left side; right side is flush to column)

- [x] Aspect ratio of both image and size of card below should be 1:2 (just like the cards above, tho since there's a gutter they will be smaller in size); image and card will be same aspect ratio and size at each breakpoint

- [x] Move author and date to the bottom of the card ($caption-01) 16px from bottom and side

- [x] Change color of author and date to BEBEBE (Gray 30)

- [x] Change article name to $body-short-02 (not semi-bold anymore)

- [x] On Start Designing / Start Developing -- use the attached images for tablet and mobile (i.e. when the card goes below the image)

<b>Spec</b>

![image](https://user-images.githubusercontent.com/43144260/51277366-0969cf80-199d-11e9-9cd9-ed100020721e.png)

[Getting_Started_imgs.zip](https://github.com/carbon-design-system/carbon-website/files/2797702/Getting_Started_imgs.zip)

PR #810, opened 2019-01-28
Homepage updates

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/724

- Adjusts margins on FeaturedTiles as well as ArticleTiles (Both have a margin on mobile now)
- New images on FeaturedTiles for mobile
- Changed some font styles

- Removed a `console.log` I found


jeanservaas:
Looking so much better -- couple things:


- Can we change the margin bottom between "Start Dev'ing" and "Start Designing" to 16px instead of 24px?

image


- Can we add some percentage of right padding to these cards too (we can kill it at mobile)

image


- Awesome that you made this change to the Resources page too, I'm just wondering if we can fix the stacking arrangement there... i.e. under Featured, why are there two cards on top and three on the bottom... also, below in articles... they should also go to three columns like on the homepage... for some reason they are staying at two

image

- I would like to give you new images for some of these... the ones under Featured are really narrow, not the right aspect ratio etc.

- Please change the color of the author / date to Gray 70 only on the Resources page since it is on a white card not a dark one (565656)

image

jeanservaas:
Right margins look great! and cards are stacking right in Resources now!

- Resource cards accidentally got a margin bottom too:

image

image

- Also, just to double check (are Author and date Gray 70? They look same color as the article title)

- Also 2, not sure why this is happening but these imgs all appear to have a light gray rule at the top of them in -- it's in Firefoxy and Chrome.

tw15egan:
@jeanservaas updated 👍

src/components/ClickableTile/clickable-tile.scss
```diff
@@ -42,59 +59,51 @@ a.tile--article.tile--dark:hover {
   border-bottom: none;
 }
 
+.tile__img span {
+  height: 100%;
+}
+
 .tile--article .tile__img img {
   width: 100%;
+  object-fit: cover;
```
alisonjoseph: I'm not sure this works in IE11, so worth double checking how things look there

jeanservaas:
@tw15egan 

I am almost ready to approve, just getting you new images for resource page and one other comment:

One last comment:
- I think we can hav right padding at Medium breakpoint and mobile too -- the only place where I would kill it it as Large (1056) breakpoint -- like you did. Think it can go a little less at mobile 25%.

image

jeanservaas:
One last thing, Mike asked that we put this guy in instead of typing guy for the hackathon image on both homepage and resources

[Uploading homepage_03_revised.png.zip…]()

PR #811, opened 2019-01-28
feat(image-component): added custom component for images

### LINKED ISSUE #683: Image Component

marijohannessen:
Closes #683 

Added in custom component `ImageComponent`. Wraps around all the images on the website.

Example:

```
<image-component cols="12" caption="I am an image">

![image-alt-text](/images/cool.png)

</image-component>
```

### Customizable:

`cols`: The image can span over either 4, 6, 8 or 12 columns
`border`: If set to true, will have an 1px solid `$ui-03` border
`fixed`: If set to `default` or `large`, it will have a max-width of either 720px or 1120px
`bg`: If set to `none` then it will not have a white background

### TODO

- [x] Right now it only specifies columns on large screens, need to figure out how this translates down on smaller screens
- [x] Trying to figure out how to remove the image from the grid (per requirement in #683), not sure if this is possible when the max-width is set in the gatsby image plugin, do you know @alisonjoseph ?


alisonjoseph:
> Trying to figure out how to remove the image from the grid (per requirement in #683), not sure if this is possible when the max-width is set in the gatsby image plugin

The only thing I can think of is that we add a width property and add the specific width the image should be, then make sure it has a max-width 100% so it scales when the browser is smaller. Not 100% sure that would work, but I think so?

marijohannessen:
> Can we make it possible to add custom classes (in case a specific image need overrides?)

Yes definitely!

marijohannessen:
Just pushed up the initial work on how the component would handle images that don't follow the grid. @shixiedesign and @jeanservaas can you guys take a look and see if it is behaving as expected? I've only tested it out on the images on accordion/style, breadcrumbs/style, and buttons/style. These all have a max-width of 720px now. I have not tested it out on the larger variant (1120px).

marijohannessen:
Deploy link: https://deploy-preview-811--carbon-website.netlify.com/

shixiedesign:
Images (on accordion breadcrumb button) are responding correctly!! Exactly what we want. Just to clarify, we still have the option to use fluid behavior on images on Elements pages, correct? Currently the images on elements pages look affected. They are still fluid but their captions are no longer styled. 

So overall images (on component style tabs & elements pages) are behaving right, just need to restore caption styling! 

Eg. below image caption line length should be 4 col wide, and fluid with browser width...let us know if this is a difficult change. Thank you Mari!! 🙏

image

shixiedesign:
I noticed the markdown syntax is `fixed = default`. What should that be when the image is fluid?

marijohannessen:
> Images (on accordion breadcrumb button) are responding correctly!! Exactly what we want. Just to clarify, we still have the option to use fluid behavior on images on Elements pages, correct? I noticed the markdown syntax is `fixed = default`. What should that be when the image is fluid?

You just wouldn't add the `fixed` option at all!

shixiedesign:
Hey @marijohannessen , sorry I noticed an issue...both `fixed="default"` and `fixed="large"` are displaying at large size (1120px). The images with `fixed="default"` flag should display at `720px` width. Is this an easy fix?

PR #813, opened 2019-01-28
Edits to Theming page

claycrenshaw:
General cleanup

Added index links

New instructions on setting up a new theme

**Note: need confirmation that the actual token data is still accurate for new v10 token system.


jeanservaas:
Just a couple of comments

As we already discussed, just wanna make this consistent now with the flow outlined in this issue:
https://github.com/carbon-design-system/carbon-website/issues/831

- If there's a resource tile, give it a title, add it to the index and treat it the same as all the other ones

- also, if there are two resource tiles, put them together at the top of the page

- Could we loose the word "Introduction" ... maybe something instead like "Theming you UI" "Understanding themes" "Applying themes" etc

image

- Is this code snippet supposed to be in all caps? Just checking.

image

aagonzales:
Good point Jeannie about moving the resources tiles to the top for consistency and also adding in the GitHub one. Make sure it links to the elements[ theme package ](https://github.com/IBM/carbon-elements/tree/master/packages/themes)and not the old `carbon-themes` repo.

Only other thing is the "Carbon default theme" sections:
- I think we can clarify that the default theme is the "White theme".
- remove the {, a.k.a. “Duo.” } part. I don't think we're supposed to be referencing Duo in our official docs. 
- Make section and H3 then (like Jeannie suggestion)

claycrenshaw:
Replacing this with a clean PR.

Issue #762, opened 2019-01-21
[Global]: The italics are forced italics rather than Plex italic fonts.


mjabbink:
This seems to be a global issue but you can catch it in the Patterns, Accessibility pages and wherever you might fins italics being used. https://next.carbondesignsystem.com/patterns/overview

You may have resolved already.

mjabbink:
<img width="698" alt="screen shot 2019-01-20 at 12 40 55 pm" src="https://user-images.githubusercontent.com/32881239/51485391-ec9e1500-1d6b-11e9-8cf3-18e978905551.png">
<img width="646" alt="screen shot 2019-01-20 at 12 40 50 pm" src="https://user-images.githubusercontent.com/32881239/51485392-ec9e1500-1d6b-11e9-9e4b-57e4ffeb91a3.png">
<img width="719" alt="screen shot 2019-01-20 at 12 40 23 pm" src="https://user-images.githubusercontent.com/32881239/51485393-ec9e1500-1d6b-11e9-9a1b-584029a57dac.png">
<img width="740" alt="screen shot 2019-01-20 at 12 40 05 pm" src="https://user-images.githubusercontent.com/32881239/51485394-ec9e1500-1d6b-11e9-8e8b-8a7f591dbc7f.png">

mjabbink:
<img width="858" alt="screen shot 2019-01-21 at 12 08 41 pm" src="https://user-images.githubusercontent.com/32881239/51489097-5838b000-1d75-11e9-98a7-a54ff7dc1cf1.png">

PR #815, opened 2019-01-28
chore(elements): update elements for font hotfix

joshblack:
Closes https://github.com/carbon-design-system/carbon-website/issues/762

Update elements dependencies for font hotfix in: https://github.com/IBM/carbon-elements/pull/301

#### Changelog

**New**

**Changed**

- Elements packages are now on `alpha.31`

**Removed**


joshblack:
cc @mjabbink if you have a chance to verify that italics are no longer the forced variants 👍

alisonjoseph:
Italic text looks identical on live site and preview link. 
screen shot 2019-01-28 at 6 34 24 pm

mjabbink:
those italics in above screen grab comparison are both forced italics

mjabbink:
Also on Typography page in this section:

screen shot 2019-01-28 at 8 19 07 pm

alisonjoseph:
@joshblack I'm not setting the font family anywhere, is this because I'm still pulling in v9 Carbon experimental and not v10 alpha?

joshblack:
@alisonjoseph yeah, I believe the reset from type sets the font family but it's most likely being overwritten by the carbon styles.

https://github.com/IBM/carbon-elements/blob/master/packages/type/scss/_reset.scss#L18-L25 in the reset, but we could totally isolate the styles using the `font-family` mixin so that the font families match: https://github.com/IBM/carbon-elements/blob/master/packages/type/scss/_font-family.scss#L16

alisonjoseph:
fyi 🎉 
screen shot 2019-01-29 at 3 34 08 pm

mjabbink:
yes!

mjabbink:
:muscle-ltr-270:

Issue #663, opened 2019-01-09
Develop 404 page


shixiedesign:
Develop 404 page. Design provided in issue #631 by @shixiedesign 

This should be built as a website-addon.

PR #855, opened 2019-01-31
404 page

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/663

Adds in custom 404 page.

The repeating background is skipped for now and added an animation on the gradient. 

We still need to figure out why the site will not redirect to the 404 page, works fine locally.

To test, go to `/404`

screen shot 2019-01-31 at 1 29 38 pm


tw15egan:
Looks like redirect is working, but now the sidebar color change isn't working. Will update and find a way to tell the nav bar if it is on the 404 page

shixiedesign:
Hover of links now have to be `Blue 40` because it's against a dark background. Gotta make sure this local change won't affect the links on regular pages though. 
Current:
image

alisonjoseph:
Can we pull this out in to the website-addons so I can use inn the ibm.com/design site?

claycrenshaw:
Looks good! Could I suggest a tweak to the language?

IBM style says not to use exclamation points for negative messages. . . But we can have some fun with it, and make it a little more conversational. Maybe like. . .

Something's wrong here :(
Sorry, we can't find the page you're looking for. But maybe one of our most-visited pages will suit your fancy?

tw15egan:
@alisonjoseph added!

alisonjoseph:
🎉 thanks! looks good.

alisonjoseph:
@tw15egan I started digging in to why the 404 page wasn't working and got as far as it having something to do with `<Layout>`. If you remove Layout then it works. No idea how to fix it, found a bunch of older gatsby issues related to it, but didn't really dig in to them. (you can test by removing layout, running `yarn build:external` and then `gatsby:serve` and checking a random url. (also confirmed identical behavior on /design site)

Obviously this is not ideal since we want the navigation on our 404 page, however we might have to do this as a temp fix for next week. 😭 

issuess possibly related? 😕 
https://github.com/gatsbyjs/gatsby/issues/2223
https://github.com/gatsbyjs/gatsby/issues/1838

alisonjoseph:
never mind! update Gatsby to `2.0.107` and I think it fixes the issue 🙏

alisonjoseph:
oh, this wasn't ready to merge. (I'll update gatsby in a separate PR though to actually make the 404 page work and publish the package)

PR #863, opened 2019-02-01
fix(content-switcher): added gray bar at the top

### LINKED ISSUE #757: Color Token component (background tweak)
## Feature request: Color Token component (background tweak) 

> I missed this during revisions, so sorry, but mike called it out: could the color token component disppear under the content switcher and not appear again? Right now, you can see it scrolling in the gap between the switcher and the browser. Sorry!

- also, this is a bug in safari... rules are displaced for some reason.

image

marijohannessen:
Closes #757 

Added gray bar to the top of the content switcher so you can't see the table below it when scrolling.


marijohannessen:
Also fixed bug with Safari (mentioned in the same issue)!

jeanservaas:
awesome thanks!

Issue #249, opened 2018-10-10
[Web Component] Icon card & search/filter bar


designertyler:
Even though our new site is extremely fast, the new experimental iconography page still triggers a full page loading spinner overlay.

Is there a better way to handle this? Can we use skeleton states? Or even just remove the spinner overlay altogether?

![Screen Shot 2018-10-10 at 11.14.55 AM.png](https://images.zenhubusercontent.com/583dd918b9dc3c622022afbf/d0841aec-37ee-4d43-8bad-9bd0beb88270)

joshblack:
@designertyler this detail was added since we had to switch to loading more icons. This was the stop-gap to help handle folks viewing this page on slower connections while we figured out a better way  👍 

From the technical perspective, can 100% implement skeleton states or anything else that is preferred!

alisonjoseph:
Yep like Josh said, this was a first pass just to get all the new icons up there for people to view. There are SO many icons now might be worth exploring different ways to display/organize them.

designertyler:
Thanks for the quick responses. This isn't a priority and only happens for a split second for me. A minor tweak could be to not overlay (and appear disabled) the nav and header. 

I'll create another issue for design to find a better way to organize/sort/filter the icons. That page gets pretty overwhelming.

chrisconnors-ibm:
prior to nav changes can we lazyload/skeleton state these? maybe by row

alisonjoseph:
Yea we could do skeleton state loading just depends on where this falls in our priority list. @joshblack how much effort do you think that would be?

joshblack:
Do we have designs for the skeleton states? If they are minimal, shouldn't be too bad.

shixiedesign:
Beta release nice-to-have: 
Icon card skeleton state. Making this part of the to-do under **[Web Component] Icon card**

shixiedesign:
Icon Card spec
Sketchfile: https://ibm.box.com/s/o5qhcg064uzwedl63yimxh1hzx37kcn7
![image](https://user-images.githubusercontent.com/15144993/49815986-91234500-fd32-11e8-829e-0025f31f1eaf.png)

shixiedesign:
# Final icon card design and spec
Sketchfile: https://ibm.box.com/s/o5qhcg064uzwedl63yimxh1hzx37kcn7
![image](https://user-images.githubusercontent.com/15144993/50427111-b5b80a80-0864-11e9-9d05-9f0127132b2f.png)
[...]

PR #874, opened 2019-02-01
Icon page update

shixiedesign:
Address the 2 fixes mentioned in https://github.com/carbon-design-system/carbon-website/issues/249

- [x] Take outside stroke off the component 
- [x] Remove repeated icon name

Sorry I'm not good with css and my solution to remove outside stroke is probably not the most elegant. Because the number of cards per row changes across browser widths (and not by breakpoints) I had a bunch of `@media` queries. Please let me know if you got better fixes! Thank you!  🙏


asudoh:
Good question @shixiedesign - I'd seek for "overriding larger border by layer" approach, though I guess your approach is fine for the interim.

shixiedesign:
Hi @alisonjoseph I was actually trying to use the `@include breakpoint('lg')` but it doesn't work because the number of cards per row is not responding to breakpoints (they are fixed, not fluid), so the number of cards changes in-between breakpoints too. Would happy to pass this on to a dev if anyone has bandwidth.

PR #891, opened 2019-02-05
fix: switch addons-website to use alpha over latest release

vpicone:
The release tag for `@carbon/addons-website` doesn't include components required for dev. Yarn doesn't use semver for alpha packages so it's effectively pinned to `alpha.0`.

Sidebar: maybe using the `alpha` tag for the carbon-elements and alpha Carbon packages could be a better solution than pinning them to `alpha.31`

Came across this issue when doing a regular install → dev command. The build fails when importing `FourOhFour` in `/src/pages/404.js`


vpicone:
I need a tool to that plays the yakkity sax whenever I make a PR with offline-mirror issues.

Issue #881, opened 2019-02-04
Update v9/v10 banners


alisonjoseph:
Add clear documentation to the website around v9/v10 components with a link to the v9 site. The v10 website is launching before we release v10 carbon-components so this is super important.

aagonzales:
The banners are slightly different for each site. Used the v9 button for the v9 site and a v10 button for the v10 site. Hope that's ok/preferred.

bar height: 48px
text: 14px, regular

Sketch file: https://ibm.box.com/s/pxjarp8zdpqlyw6alek26ucqq3e6iq9g

## v9 site
![image](https://user-images.githubusercontent.com/11670886/52238023-a7f89a80-2890-11e9-9a42-1688faa9e1a5.png)


## v10 site
![image](https://user-images.githubusercontent.com/11670886/52237997-98795180-2890-11e9-8b39-ef39f320e3f0.png)

claycrenshaw:
for v9: 

"Carbon X has arrived! Check out the v10 beta site and all-new carbon-components library."
CTA: "Go to beta site"
(rationale: "is here" at first glance could be read as "is here, on the page you are reading")

for v10:

"Carbon v10 is currently in beta; for production-ready components, use Carbon v9."
CTA: "Go to v9 site"

Just a note: esp. on the new site, the button on the banner is so far from the related text that it seems like it would be easy to miss it. . . Also not sure if there is an easy fix for this. Possibly just make the banner text clickable as well?

shixiedesign:
Talked to Anna about V10 site banner. The white is merging into the browser controls above it. So we made a new design...many many apologies for this late design change. If it makes it that will be awesome, it not it's not the end of the world. 🙏 thanks!!! 

Top line is the design. Bottom line is spec'ed design.

![image](https://user-images.githubusercontent.com/15144993/52362618-989b5d80-2a06-11e9-958f-e4f501219b95.png)

alisonjoseph:
- [ ] Need to update banner on v9/v10 site to not be sticky
- [ ] Update banner on v10 site to use updated design

alisonjoseph:
Can/should we keep the banner on the next site at 48px high on mobile also now that we are hiding text and it will fit at that height?

alisonjoseph:
Fix is in this PR https://github.com/carbon-design-system/carbon-website/pull/917

PR #894, opened 2019-02-05
feat: add banner to v9 website

alisonjoseph:
Ref #881 & PR https://github.com/carbon-design-system/carbon-website/pull/893

@aagonzales same question here around mobile

### DIFF
--- a/src/components/Layouts/index.js
+++ b/src/components/Layouts/index.js
@@ -144,14 +144,14 @@ class Layout extends React.Component {
 
             <div className="website-alert">
               <p className="website-alert__text">
-                Carbon X is here! Check out the new version 10.0.0-beta site and
-                carbon-components.
+                Carbon X has arrived! Check out the v10 beta site and all-new
+                carbon-components library.
               </p>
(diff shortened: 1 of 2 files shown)


alisonjoseph:
Updated, the text seems a little bit long on mobile though. 
@claycrenshaw suggestions to shorten the text?

screen shot 2019-02-05 at 8 17 23 pm

claycrenshaw:
For mobile I say we can just kill the second sentence. Also please be sure to include the hyphen in "production-ready". Thanks!

alisonjoseph:
Can we get copy that works across all breakpoints. Also "Carbon X has arrived" might not be the best thing to say, since it hasn't actually arrived yet 😉

aagonzales:
What about  just "See what's coming soon with Carbon V10-beta."

aagonzales:
And can we make the banner not sticky and go away on scroll down?

alisonjoseph:
Copy updated. will look at making it not sticky in separate PRs.

PR #895, opened 2019-02-05
feat: add global search

### LINKED ISSUE #494: [Website Component] Global Search
Need design spec for search box and algolia results box

fyi #299 & #298

aledavila:
Closes #494 

#### Changelog

**New**

- global header search input

**Changed**

- algolia search box changes

**MIGHT NEED TO UPDATE COLORS TO TOKENS**


src/components/GlobalSearch/GlobalSearch.js
```diff
@@ -37,7 +36,11 @@ export default class GlobalSearch extends React.Component {
           id="doc-search"
           placeholder="Search"
         />
+        <Close20
```
tw15egan: I think this may need to be wrapped in a button... I'm having issues trying to close the search when clicking on the icon. Making it a button will also make it keyboard accessible, so a keyboard user can tab to the close and close it that way

src/components/Layouts/index.js
```diff
@@ -45,6 +47,28 @@ class Layout extends React.Component {
...
+    this.setState({
+      [isSearchOpen]: !this.state.isSearchOpen
+    }, this.handleSearchEventListener);
+  }
+
+  handleSearchEventListener = () => {
+    if(this.state.isSearchOpen) {
+      document.body.addEventListener('click',this.handleCloseSearchClick);
+    } else {
+      document.body.removeEventListener('click',this.handleCloseSearchClick);
```
tw15egan: Do we need to remove the eventListener?

src/components/Layouts/index.js
```diff
@@ -45,6 +47,28 @@ class Layout extends React.Component {
     });
   }
 
+  handleSearchClick = (isSearchOpen) => {
```
tw15egan: We should call this on keypress as well

src/components/Layouts/index.js
```diff
@@ -202,6 +226,16 @@ class Layout extends React.Component {
 
               <HeaderGlobalBar>
                 {/* {isInternal ? null : <GlobalSearch />} */}
+                {this.state.isSearchOpen ? <GlobalSearch /> :
```
tw15egan: Instead of removing the entire element from the screen, can we toggle a specific class? Something like `bx--header__action--search--active` 

That way we can set the width to 0 and make its visibility hidden when closed and animate the width to 448px on click.

src/components/Layouts/index.js
asudoh: Weighing in - I think code like this is important to avoid event listener leak.

src/components/GlobalSearch/GlobalSearch.js
vpicone: Using a `blur` event on the search component would also make sure that the search closes whenever a keyboard user leaves the search in either direction. MDN has a similar component that functions this way. 

Might be an alternative to managing the click event listener on the document as well.

src/components/Layouts/index.js
aledavila: Yea I was removing the event listener as so it wasn't going off in the background on ever click outside. I will go over the component today and make revisions thanks for the input guys.

alisonjoseph:
Since we aren't replacing the old site just yet I applied for docsearch for the next site, we need to wait to merge until that is approved and I have the updated code to add. (otherwise we will get search results for www)

src/components/Layouts/index.js
tw15egan: Makes sense!

src/components/Layouts/index.js
aledavila: When I tab to the magnifying glass it open on enter. Do you mean that it should close on a keypress ?

alisonjoseph:
The global search on the v9 version of the site just broke the entire site when it went down. Can you update the version in html.js to v2.6.2 
`<script src="https://cdn.jsdelivr.net/npm/docsearch.js@2.6.2/dist/cdn/docsearch.min.js" />`

and also wrap `window.docsearch({` code in an else statement so that if it does break the entire site doesn't go down. 😅

aledavila:
OMG thats so scary yea

tw15egan:
Should we consider pulling that file down and including it on our end? That way we don't have to rely on the external script

vpicone:
@tw15egan Would it be better to just include it as a proper dependency in that case? Pinning the version in the script tag is definitely a good idea regardless.

tw15egan:
Makes sense, as long as the version is locked down I'm happy

shixiedesign:
Hi Ale, looking great! Thanks for doing this. Some little design fixes:
- [ ] the search results are missing `active` states: bg should be `Gray 30`. (My fault completely. I forgot to provide in spec.)
- [ ] hover & active background height should be 34px (8px padding above & below text)
image
- [ ] hovering category titles highlights the option underneath instead
untitled_2_1
- [ ] underline of word "lost" should take the Blue 60
image
- [ ] need add 16px gap above aloglia logo 
image

asudoh:
@aledavila Weighing in - Just wondering if @tw15egan's comment has been addressed - Thanks!

aledavila:
@asudoh yes except the way the search box appears. Hiding versus on width change

asudoh:
OK looking forward to seeing the update on the search box - Thanks @aledavila!

tw15egan:
Ship it!

PR #905, opened 2019-02-06
[Image Component] Style image updates

jeanservaas:
Some of the style images were saved out at 544px (2x pngs); this was causing a lot of extra white space within the image component:

- I went back through and resaved them at 736px (2x pngs)
- There were also a lot of instances of Plex Sans still in the specs... so I changed those to Plex Mono

It was easier to just update all of the images again because of the Mono changes rather than try to keep track of which ones I changed and which ones I didn't.


src/content/getting-started/developers/react.md
```diff
@@ -43,11 +57,6 @@ $ yarn add carbon-components-react carbon-components carbon-icons
 
 3.  For older browsers (e.g. IE11), polyfills listed in the [carbon-components-react/.storybook/polyfills.js file](https://github.com/carbon-design-system/carbon-components-react/blob/master/.storybook/polyfills.js) are required.
 
-If you just want to try out `carbon-components-react`, you can also use [CodeSandbox](https://codesandbox.io).
```
alisonjoseph: was this removed on purpose? there are a bunch of updates here that don't seem related to images?

jeanservaas:
not sure why this is in this PR... i must've done it before changing branches... ug. That change is reflected in another branch's PR.

PR #871, opened 2019-02-01
Visual sweep of Content page

jeanservaas:
Visual clean-up the content page: everything that I'm doing here is listed in detail in this issue: 

https://github.com/carbon-design-system/carbon-website/issues/792

PR #909, opened 2019-02-06
Update CONTENT.md

claycrenshaw:
Updates to MD rules


jeanservaas:
Hey @claycrenshaw, I have a Content PR too, can you view it... 

https://github.com/carbon-design-system/carbon-website/pull/871

now they're in conflict but there's a lot in mine that I don't want to lose. 

**Does yours only affect the Guidance tab?** 
The best thing to do might be to just push yours though and ditch mine, but I had a lot of comments on the Guidance tab.

- I really like the anchor links that you added to the Guidance tab

- My main work was around the Do / Don'ts captions; I think it's confusing and inconsistent with the way we treat the rest of our guidance to have the same caption under both the "Do" and the "Don't" -- you don't have to use exactly what i've written (i have examples for everything on the page in my PR) but could you at least change them to "Do: Future tense" / "Don't: Future tense" so that they say something different?

**Yours:**
image

**Mine:**
image

- Could you add a Don't for Please? and take out the title "Using Please" OR write "Do: Using please" "Don't: using please" ... just whichever we choose, be consistent ( See my example).

*Yours:**
image

**Mine:**

image

- Could you eliminate the words "you are" from both of these examples -- they will look a lot better and it won't affect the read -- i.e. "You might need more advanced features when integrating with another app."

image

- I also had some formatting changes (nothing with content) on the General tab and the Glossary tab... if you haven't changed any of that, once you push your PR through, I can make a new one just for those quick tweaks.

claycrenshaw:
@jeanservaas I think you have the wrong PR here. . . This is for the content.md file which is a readme explaining how to make updates to the website (for carbon team members)

claycrenshaw:
The preview link is misleading, easy mistake.

claycrenshaw:
@jeanservaas Thanks for this feedback. Great points. I integrated it into the updated version and will submit through a separate PR.

PR #911, opened 2019-02-06
Themes tile update

jeanservaas:
- brought themes tile up to the top for consistency but also left link to themes in the "Default theme" paragraph below.

- updated theme "terms" from bold to italic


alisonjoseph:
Looks like there are a bunch of image updates in this PR? Are they supposed to be?

PR #914, opened 2019-02-06
Small color page tweaks

jeanservaas:
- put correct text in lead space on the usage tab

- fixed spacing above themes table


alisonjoseph:
this PR also has a bunch of component images in it? When you start new changes make sure you are branching from the latest version of the master, I think maybe you are just branching off of the last branch you are working on, which means you will get both sets of updates in there.

PR #917, opened 2019-02-07
fix: update banner styles

alisonjoseph:
ref https://github.com/carbon-design-system/carbon-website/issues/881#issuecomment-461276843

banner is now 40px tall, and copy was updated


alisonjoseph:
updated @shixiedesign

PR #927, opened 2019-02-07
adding usage images

designertyler:
Added new images for notification usage and edited markdown.


src/content/components/notification/usage.md
```diff
@@ -27,6 +27,15 @@ We recommend that toast notifications automatically disappear after five seconds
...
 
+<image-component cols="8">
+
+![The three icon states](images/notification-usage-1.png)
+
+</image-component>
+
+## Placement
+
+
```
marijohannessen: It looks like there are two H2's next to each other here, is that intentional?

jeanservaas:
quick comment:

I know that one slider image is an animation and one is static, but just wondering why they're not the same height and also why one goes full screen and one is short -- I like the proportions and size of the shorter one better, wondering if they could just both look like that?

designertyler:
@jeanservaas Fixed the animated one to be 8 columns like the static image.

src/content/components/notification/usage.md
designertyler: There was a duplicate `## Placement` in the wrong spot. I deleted the extra.

PR #940, opened 2019-02-07
fix(do-dont-components): fixed padding and top line

### LINKED ISSUE #903: [Website component bug] Do / Don't component has added padding
Need to remove forced padding on Do / Don't component so that it is optional / not true by default:

- please also make top bar 3px tall (it is 2px now)

See example of forced padding causing an issue on the Color overview page:

marijohannessen:
Closes #903 

Removed padding on the `DoDontExample` component and switched from `rem` to `%` padding for the text content so it doesn't flow outside of the box. Also changed the height of the top line from 2px to 3px.


designertyler:
This is happening with the top bar
image

alisonjoseph:
I'm seeing the same issue as Tyler on typography page. I think it has to do with the default margin getting added to the bottom of images.

marijohannessen:
@alisonjoseph @designertyler Fixed the issue!

Issue #983, opened 2019-02-08
QA [Website image component] Fixed component images are malfunctioning in mobile


jeanservaas:
Component images are extending beyond the canvas in mobile:

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/e7316804-c1c6-46b9-9534-6aabd2ea1724)

tw15egan:
Seems to only be happening in Safari

tw15egan:
Seems to be resolved /shrug

PR #1005, opened 2019-02-08
fix(images): fix images on mobile in safari

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/983

Adjusts image widths on mobile

cc @marijohannessen

### DIFF
--- a/src/components/ImageComponent/image-component.scss
+++ b/src/components/ImageComponent/image-component.scss
@@ -46,9 +46,17 @@
 
   img.gatsby-resp-image-image {
     max-width: 736px;
-    width: 736px !important;
+    width: 100%;
     height: auto !important;
     position: relative !important;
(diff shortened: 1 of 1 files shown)


src/components/ImageComponent/image-component.scss
```diff
@@ -46,9 +46,17 @@
 
   img.gatsby-resp-image-image {
     max-width: 736px;
-    width: 736px !important;
+    width: 100%;
     height: auto !important;
     position: relative !important;
+
+    @include breakpoint('sm') {
+      min-width: 368px !important;
```
marijohannessen: We want to keep the images 736px as long as possible, and with this added code the image shrinks very early on. I wonder if we can instead to a min-width only on mobile screens?

Issue #943, opened 2019-02-08
Tab component missing 16px padding


shixiedesign:
Bug with motion page tabs component
- [ ] missing 16px padding all around
- [ ] missing background behind images
- [ ] clicking on a tab for some reason scrolls back to top of page

![image](https://user-images.githubusercontent.com/15144993/52457691-d9879500-2b20-11e9-8b6b-e84475c4772b.png)

lovemecomputer:
this issue addressed in https://github.com/carbon-design-system/carbon-website/pull/1010

Issue #942, opened 2019-02-07
Tab behavior bug on motion basics page


kellychurchill:
Clicking a tab causes browser to jump to top of page... on https://next.carbondesignsystem.com/guidelines/motion/basics - second set of tabs down the page has anchor links that on clicking a tab (say, the Abstract tab) go to https://next.carbondesignsystem.com/guidelines/motion/basics#top with result that visual context is lost and user has to browser down to see result of choosing the tab

lovemecomputer:
adressing temporary solution in https://github.com/carbon-design-system/carbon-website/pull/1010 

need to eventually circle back and make a longer-term solution which doesn't require an author to add `href="javascript:void 0"` to every `<tab>` component

lovemecomputer:
this issue has been closed, just watned to update that the *actual* fix was put in in that PR linked above!

PR #1010, opened 2019-02-08
WebsiteTabs fixes

lovemecomputer:
Closes https://github.com/carbon-design-system/carbon-website/issues/943

temporary solution for https://github.com/carbon-design-system/carbon-website/issues/942 - need to eventually circle back and make a longer-term solution which doesn't require an author to add `href="javascript:void 0"` to every `<tab>` component. thanks @asudoh !


src/components/WebsiteTabs/WebsiteTabs.js
```diff
@@ -49,6 +49,8 @@ export default class WebsiteTabs extends React.Component {
   render() {
     this.updateTabChildren();
 
+    console.log(this.tabChildren[0].props);
```
dakahn: left a log in

tw15egan:
Looks like this can be turned off from the package: https://www.npmjs.com/package/smooth-scroll

Just need to set `topOnEmptyHash: false` 

screen shot 2019-02-08 at 4 26 27 pm

src/components/WebsiteTabs/WebsiteTabs.js
lovemecomputer: @dakahn thanks!

lovemecomputer:
thanks @tw15egan ! looks like all is good with this PR now~

PR #1041, opened 2019-02-08
Color tokens update

aagonzales:
- Added missing tokens 
- Updated a few wrong usages


aagonzales:
Oops not sure why my QA branch is also showing up 🤷‍♂️

PR #1055, opened 2019-02-09
fix(ComponentExample): fix notification in narrow width

### LINKED ISSUE #1002: [Notification]: code tab, component example mobile bug - Chrome
## Detailed description

> Describe in detail the issue you're having.

In mobile view, the notification text is running off of the actual notification box.
In desktop view, it seems like the text is also too close to the edge, needs more padding.
This might be just because it's in the demo container.

> Is this issue related to a specific component?

Notification

> What browser are you working in?

Chrome (dev tools to view mobile)
notification code-mobile
notification code

asudoh:
Fixes #1002.

#### Changelog

**Removed**

- Override style for notification width, causing the issue. Couldn't find a negative impact by removing it.


shixiedesign:
The text running out of bound is fixed! Thanks! Looks like two notifications now are displaying on the same line but not exactly aligned. Can we have them displayed in separate lines, or somehow align them horizontally? 

image

asudoh:
Good catch @shixiedesign! Fixed.

PR #1065, opened 2019-02-11
Update search image (usage)

laurenmrice:
I don't know what the conflict is, I just wanted to override the image.?


laurenmrice:
Yes that looks correct 👍

Issue #803, opened 2019-01-28
Add React component examples for Dropdown to website


alisonjoseph:
Add React examples to website for Dropdown. (see current site for examples https://www.carbondesignsystem.com/components/dropdown/code) Could use codesandbox plugin for this. 

- [x] Multiselect Dropdown
- [x] Inline Multiselect Dropdown
- [x] Filter Dropdown

joshblack:
- MultiSelect: https://codesandbox.io/s/50739oq2lx
- Inline MultiSelect: https://codesandbox.io/s/zq9o7l4543
- Filterable MultiSelect: https://codesandbox.io/s/ovx9j0rkq9
- Combobox: https://codesandbox.io/s/kkrnjj5vx7

joshblack:
cc @mattrosno any tips for using these in Gatsby with the codesandbox plugin?

mattrosno:
@alisonjoseph @joshblack just seeing this. I'll submit a PR to get working codesandboxes in there, separate from tutorial PR.

Not sure how good the embedded codesandboxes would look in there... sure we don't want to simply link out to React Storybook like how it's handled in v9 components?

alisonjoseph:
Oh, for examples above, should we switch to experimental v10 versions for the next site?

asudoh:
@alisonjoseph Good point - Created #1057 for that.

asudoh:
Oh sorry I had a wrong assumption that we put the React components directly to `<ComponentExample>`.

asudoh:
Created experimental styled version at: #1058

asudoh:
Created #1059 just in case more direct live demo (like vanilla ones) is more desirable.

PR #1074, opened 2019-02-11
fix(codesandbox): use fixed version of gatsby-remark-embedded-codesan…

mattrosno:
…dbox

Ref #803.

While waiting on https://github.com/elboman/gatsby-remark-embedded-codesandbox/pull/6 to be approved and published, install a forked version that includes the fix.

If we don't want package.json to reference a user repo, we could fork gatsby-remark-embedded-codesandbox from the carbon-design-system org?


alisonjoseph:
Do you still need this for the tutorial work you are doing? We ended up not needing it for the React examples

mattrosno:
@alisonjoseph I'll still need this for tutorial work, but if you don't need this now, you can close this PR and I can include it in tutorial PR. Your call.

alisonjoseph:
👍 you can go ahead and include it in that one. thanks!

PR #1102, opened 2019-02-12
fix(video): fixed the margins on the video component

### LINKED ISSUE #1000: Extra margin added near videos
It seems there's extra top or bottom margin added to the video component. 
screen shot 2019-02-08 at 12 55 21 pm

marijohannessen:
Closes #1000 

Fixed the height of the iframes on bigger screens so they don't have so much space on the top and bottom.


alisonjoseph:
The weird % has to do with css trickery to get aspect ratio's to scale correctly. https://www.w3schools.com/howto/howto_css_aspect_ratio.asp

PR #1107, opened 2019-02-13
Update color-tokens.js

aagonzales:
**Change:**
- Table was missing token `$ui-background`


aagonzales:
Closing without merging because Shixie is going to make change in another open branch with some other updates.

Issue #991, opened 2019-02-08
Glossary side nav needs to be sticky earlier


marijohannessen:
The side nav on the Glossary page (the A, B, C, ...) needs to be sticky earlier. It does not get sticky until halfway into B at the moment.

PR #1125, opened 2019-02-14
Glossary page style updates

### LINKED ISSUE #569: [Website Component] ABC..XYZ index for Glossary page
ABC..XYZ index on right side of Glossary page. Need design for this component. May be as simple as a color change.

image

alisonjoseph:
closes #569 
ref #991


tw15egan:
What issue is this closing? #1125 seems to be this PR 😄

alisonjoseph:
@tw15egan oops #569

tw15egan:
Looks good for the most part, just wondering how we should handle it when you click on `W`

screen shot 2019-02-14 at 4 10 16 pm


Should it have a background? Stick above?

alisonjoseph:
Hmm, good point, on our current site it just goes below the footer. What if the window is smaller than the alphabet list, should it not be sticky? 🤔

tw15egan:
Whatever is easiest, just want to make sure it doesn't overlap

alisonjoseph:
updated @tw15egan the nav now won't be sticky if the window height is smaller than the height of the nav, and when you scroll to the bottom it gets hidden behind the footer. Still not ideal, but it looks like Mike wants to change this to a top sticky nav at some point anyway.

shixiedesign:
Looks like alphabet index need to move up a bit more to make all alphabet visible. Right now, on laptop screen, I have no way of clicking on "w": 
image

lovemecomputer:
would it make sense to make the glossary side tab scrollable until you reach the bottom of the list of letters (plus padding)?

or, set vertical flex on the letters so they all fit in the space above the back-to-top button?

jeanservaas:
Hey @alisonjoseph 

The side component looks great to me at first glance... @shixiedesign have you looked? 

I had a couple of comments about the type hierarchy -- I wanted to tweak it in markdown last week but the glossary content looked like it was nested somewhere where I couldn't find.

I just wanted to fix the hierarchy, especially in light of the accessibility comments we got yesterday... Common Actions and Idioms should not be H2s everywhere on the page... it's very repetitive. This is how I'd like to style it below... **if you can show me how to get to the markdown I can do it too:**

There are only two things I'd need dev help with: 

- I wanted to pull in a token $display-02 token for the letters "Aa, Bb ..." 

- The spacing between the sections is looking like a lot more than 80px (i.e. before Common Actions and before Idioms looks huge) -- see grab below spec

glossary_hierarchy


<br>

**too big margin top:**
image

marijohannessen:
I like @lovemecomputer 's suggestion about making it scrollable as we can't really only design it for a 15" laptop screen when someone might view it on 13" or 12" as well. Everything else looks good though! 👍

shixiedesign:
@marijohannessen @lovemecomputer true...I agree with yall. @alisonjoseph is that a possible add? I think taking the letters down to 24 is still good. It was too roomy before. If we can add scrolling, so "z" sits above back to top, we can ensure all browser sizes can access all letters.

alisonjoseph:
Yep, I'll see if I can update to fit/scroll

alisonjoseph:
@jeanservaas this page isn't rendered with markdown, I'll update the type styles too

alisonjoseph:
updated the typography @jeanservaas 

still trying to work out a way to make the nav sticky, and have a scroll bar, and figure out the height it should be based on the window.

alisonjoseph:
Hey @marijohannessen @lovemecomputer do either of you have an idea on how to make the side nav behave as suggested? This is what's happening right now

If the abc nav is tallter than the window, it isn't sticky and the page just scrolls as normal (we could also just hide the nav for now at these smaller sizes, we hide it on mobile)

If the nav does fit, it becomes sticky when you scroll down to that part of the screen. Still behaves weird because the footer is GIANT so it goes behind it. 

Not sure how to have the nav with a fixed position from the top, while also being scrollable and responding to the footer. 

Wondering how much time it's worth spending on this since it sounds like its getting redesigned soon anyway.  Unless someone can point me in the right direction, maybe I'm missing some easy solution. 😅

jeanservaas:
thank you so much for doing that @alisonjoseph -- it's just much cleaner this way, quick fix.

one last thing gulp, (don't think i'm an Aa-hole!!!) can you add a lowercase letter next to the Uppercase
?
i.e. 
(Aa, Bb, Cc, Dd...)

alisonjoseph:
done @jeanservaas looks so much nicer with those text updates 👍

tw15egan:
I'm fine with how this is now, I think it's okay for now if it's simply hidden on mobile. I think updating the styles and not being broken is more important, especially if this is going to get redesigned.

marijohannessen:
@alisonjoseph I'm fine with how it is as well, especially since it's being redesigned! 👍

PR #1137, opened 2019-02-15
adjusted vectors to correct sizing and grid placement

chrisconnors-ibm:
#### Changelog

**New**

- adjusted vectors to correct sizing and grid placement for the file `iconography_usage_sizing-alt.svg`

**Changed**

existing on the left, new asset on the right.

screen shot 2019-02-15 at 4 08 38 pm


chrisconnors-ibm:
okay either Illustrator is off for presidents day, or gh, or our optimizer but this isn't what I checked in:
image

should be:
image

chrisconnors-ibm:
let's check again after this deploy.

chrisconnors-ibm:
ok I think this looks gtg

Issue #750, opened 2019-01-18
Image caption length is short and creates awkward line breaks


designertyler:
The caption length seems short and causes awkward line breaks.
Needs design spec for column width.

![Screen Shot 2019-01-18 at 9.57.10 AM.png](https://images.zenhubusercontent.com/583dd918b9dc3c622022afbf/9acd8a0e-a046-4a84-baf2-d1336740dda5)

shixiedesign:
This is awkward...though we might wanna look into a content edit to quickly resolve this for now. How about removing "for Accordion" so it's just:
> Structure and spacing measurements | px / rem

shixiedesign:
That's the exact reason I think changing width of caption is not best option because it will inevitably create problem somewhere else. I think the reason caption width is short is to encourage short copy, is that right @jeanservaas?

designertyler:
It happens pretty frequently. Rewriting all the captions to be shorter could sacrifice some of the clarity in guidance and seems like an unnecessary limitation. 

Just clicking around I found a handful of examples...
<img width="1012" alt="screen shot 2019-01-22 at 10 06 58 am" src="https://user-images.githubusercontent.com/19270502/51548443-ef613e80-1e2d-11e9-807a-5afe197f01cc.png">
<img width="871" alt="screen shot 2019-01-22 at 10 07 05 am" src="https://user-images.githubusercontent.com/19270502/51548444-eff9d500-1e2d-11e9-9041-b9a51828803d.png">
<img width="846" alt="screen shot 2019-01-22 at 10 07 18 am" src="https://user-images.githubusercontent.com/19270502/51548446-f0926b80-1e2d-11e9-83b6-53f4a2b60a81.png">
<img width="900" alt="screen shot 2019-01-22 at 10 07 53 am" src="https://user-images.githubusercontent.com/19270502/51548447-f0926b80-1e2d-11e9-8777-55286125aeee.png">
<img width="847" alt="screen shot 2019-01-22 at 10 08 09 am" src="https://user-images.githubusercontent.com/19270502/51548449-f0926b80-1e2d-11e9-9889-d108aa3552cf.png">
<img width="857" alt="screen shot 2019-01-22 at 10 08 14 am" src="https://user-images.githubusercontent.com/19270502/51548450-f0926b80-1e2d-11e9-8253-f99156b87b6c.png">

designertyler:
[...]

PR #1146, opened 2019-02-18
Increased column width for image captions

designertyler:
Increase column width for image caption. See issue https://github.com/carbon-design-system/carbon-website/issues/750


jeanservaas:
6 columns is too long for the caption and goes against line length guidance on IDL.

The idea in keeping them short is to keep the writing short and succinct. 

Let's run this by Mike on wednesday.

image

PR #1162, opened 2019-02-19
Getting Started: Designers - language edits

### LINKED ISSUE #930: [Content] Language check of *Getting Started Designer page*
I only touched the top section. See screenshot. Other content on page is original. 
https://next.carbondesignsystem.com/getting-started/designers

claycrenshaw:
Closes #930

### DIFF
--- a/src/content/getting-started/designers/designers.md
+++ b/src/content/getting-started/designers/designers.md
@@ -24,7 +24,7 @@ title: Designers
 <anchor-links>
 <ul>
     <li><a href="#carbon-design-kit">Carbon Design Kit</a></li>
-    <li><a href="#how-to-get-it">How to get it</a></li>
+    <li><a href="#getting-the-kit">Getting the kit</a></li>
     <li><a href="#contribute">Contribute</a></li>
 </ul>
 </anchor-links>
@@ -37,9 +37,13 @@ The Carbon Design Kit is a constantly evolving Sketch document containing all of
 
 The Carbon core team releases a major update to the system approximately twice a year, including a revised design kit and new versions of the components when needed. Style refinements and bug fixes are released as continuous updates throughout the year. All design kit release updates are tracked in the [change log](https://github.com/carbon-design-system/carbon-design-kit/releases).
 
-## How to get it
+## Getting the kit
 
-There are three ways to obtain the kit. For timely updates, install the kit as Sketch Shared Library. _Please note: this method requires Sketch version 53 or above._
+There are three ways to obtain the kit; via Sketch shared library, via Box Drive, and via GitHub. 
+
+### Sketch shared library (preferred) 
+
+For timely updates, install the kit as Sketch shared library. _Note: this method requires Sketch version 53 or above._
 
 <grid-wrapper col_lg="8" flex="true" bleed="true">
 <clickable-tile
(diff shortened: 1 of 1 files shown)


src/content/getting-started/designers/designers.md
```diff
@@ -25,58 +25,57 @@ title: Designers
...
-### Release Schedule
+### Release schedule
 
-Every six months the Carbon team releases a major update to the system, including a new design kit, along with new version of the components. Style refinement and bug fixes are released as continuous updates that happen throughout the year. We track all release updates in our [change log](https://github.com/carbon-design-system/carbon-design-kit/releases).
+The Carbon core team releases a major update to the system approximately twice a year, including a revised design kit and new versions of the components when needed. Style refinements and bug fixes are released as continuous updates throughout the year. All design kit release updates are tracked in the [change log](https://github.com/carbon-design-system/carbon-design-kit/releases).
 
 ## How to get it
 
-There are three ways to obtain the kit. For timely updates, install the kit as Sketch Shared Library. Please note this requires Sketch version 53 or above.
+There are three ways to obtain the kit. For timely updates, install the kit as Sketch Shared Library. _Please note: this method requires Sketch version 53 or above._
```
aagonzales: This version number will get out of date quickly, I think just saying "requires the latest version of Sketch" will suffice.

src/content/getting-started/designers/designers.md
```diff
@@ -25,58 +25,57 @@ title: Designers
...
     href="sketch://add-library/cloud/JaVzz"
     type="resource"
      >
-    <img src="images/sketch-icon.png" alt="Carbon design kit Sketch Shared Library"  />
+    <img src="images/sketch-icon.png" alt="Carbon Design Kit Sketch Shared Library"  />
 </clickable-tile>
 </grid-wrapper>
 
-If you might work with a proprietary add-on library and still want to be connected for updates, please set up the kit via the Box Drive by following [these instructions](#get-the-kit-via-box).
+If you work with a proprietary add-on library and still want to be connected for updates, set up the kit via Box Drive by following [these instructions](#get-the-kit-via-box). (Internal IBM users only)
```
aagonzales: This section is a little off. The tile actually takes a user to the Sketch Cloud download/sync thing that ANYONE can use, whether and IBMer or not. Its a beta function right now just because we haven't done a lot testing with it. Jeoff will like this to be the preferred method of connecting to the library once we test it a little more.

Connecting through Box Drive is only for internal IBMers.

I don't think we should say "If you would rather not receive library updates," I think rephrasing to something like "For one time downloads of the kit, ...". The catch with GitHub is you have to keep going back to it to get the updates, the updates are there they're just not automatic.

src/content/getting-started/designers/designers.md
claycrenshaw: I see where you're coming from here, but the requirement is technically version 53, not the newest version of Sketch. E.g., if the latest version is v54, v53 will still suffice for this method, right? If someone is stuck with v53, they can still use this method.

src/content/getting-started/designers/designers.md
claycrenshaw: I'm not totally clear on what needs to be changed here. The tile goes with the first paragraph, maybe that is the confusion? And the Box Drive option already stipulates IBM internal only. . . 

The other note makes sense, I can change that. :)

src/content/getting-started/designers/designers.md
aagonzales: Its on Sketch 53 until Sketch 54 comes out then the requirement will be Sketch 54. They do updates on a pretty regular basis.

src/content/getting-started/designers/designers.md
aagonzales: It wasn't very clear to me that that was the first way and the tile was associated with the content above it and not below it. The other two get a section of its own where as this one is wrapped up in the summary section.

src/content/getting-started/designers/designers.md
aagonzales: Actually, on second thought this note applies at all of the methods not just the Shared Library.

claycrenshaw: OK, I must be confused about how this requirement works. . . Is it that the user's version of Sketch must match the version owned by the creator of the master file? Or is 53 the version where this capability was introduced? Thanks :)

src/content/getting-started/designers/designers.md
claycrenshaw: Yeah, I think in this case the tile might add more confusion than value to the flow of the page. Going to just change that to a link for clarity. Will also look at how those paragraphs are broken down.

src/content/getting-started/designers/designers.md
aagonzales: The user's version must match the version the master kit was made on. I'm not sure if its just major version or if it has to match minor versions as well to be compatible.

tw15egan: I get these errors pretty often (not just with 53), so I think just sticking to `latest` makes the most sense, as updating my sketch to match the kits version usually resolves it

alisonjoseph:
this should be good to go once conflicts are fixed

src/content/getting-started/designers/designers.md
claycrenshaw: Yep, we decided to make that change. Just updated.

alisonjoseph:
replaced by https://github.com/carbon-design-system/carbon-website/pull/1182

Issue #1154, opened 2019-02-19
File updloader style image is missing units


designertyler:
Add px and rem to one measurement in this image
https://deploy-preview-1146--carbon-website.netlify.com/components/file-uploader/style

![image.png](https://images.zenhubusercontent.com/583dd918b9dc3c622022afbf/83cf95df-d97d-4988-b601-331f2e0db1fe)

designertyler:
Merged!

Issue #1015, opened 2019-02-08
Artifacts on PNG's with white masks


designertyler:
Images that use white shapes to crop or mask UI will show up with glitched artifacts.

Appears in multiple images. Check the sketch file and correct any images that use white shapes as a mask. 

<img width="982" alt="screen shot 2019-02-08 at 2 14 04 pm" src="https://user-images.githubusercontent.com/19270502/52503821-354b3000-2bac-11e9-92d4-003a45cd8729.png">
<img width="419" alt="screen shot 2019-02-08 at 2 17 46 pm" src="https://user-images.githubusercontent.com/19270502/52503909-72afbd80-2bac-11e9-9fc3-855d7cee3def.png">

PR #1168, opened 2019-02-20
fix(images): component image artifacts

designertyler:
Closes #https://github.com/carbon-design-system/carbon-website/issues/1015
and https://github.com/carbon-design-system/carbon-website/issues/1154

- Removed some of the weird artifacts that showed up on cropped images.
- Added missing images to tab usage
- Added units to file uploader image


tw15egan:
Still seeing it on this image:

screen shot 2019-02-20 at 11 40 13 am

PR #1177, opened 2019-02-21
update(content):productive type specs

jeanservaas:
- Changed the name from $expressive-heading-03 to $productive-heading-03

- fixed type hierarchy

- added a bit of content


aagonzales:
And yes @alisonjoseph it was removed on purpose. Caption should only be used in the expressive set. Product uses label for similar instances.

Issue #1159, opened 2019-02-19
fix(website component: do-don't): caption needs right padding


jeanservaas:
Padding has been added but there's too much padding there now:

- [x] add 25% right padding to captions at all breakpoints @tw15egan did this for us with article component

- [ ] Check on the amount of padding that was added (see image below, it looks very extreme now).

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/81bf41f8-5d35-48d2-aaa7-9373ac22cb39)

See how IDL captions have right padding and the line breaks sooner:
![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/8d2932e2-cf54-4c3d-b2b7-6197b92d8d7a)

jeanservaas:
@tw15egan 
I noticed #1161 added the padding but I think it added too much ... can you check possibly how much was added? It should only be like 20-25%. Right now the caption is barely spanning half of the image.

![image](https://user-images.githubusercontent.com/43144260/53123616-1723ef00-351f-11e9-8117-cd686b32310d.png)

PR #1178, opened 2019-02-21
fix(max-width): remove max-width from captions, rely on padding

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1159 (for real this time)

Removes max-width, rely on padding instead


designertyler:
LGTM!
screen shot 2019-02-22 at 12 36 55 pm

PR #1225, opened 2019-02-27
Fix contribution link on home page and footer

designertyler:
The "start contributing" link on the home page was going to a 404.

The link to contribute in the footer was only going to Designer contribution.

I pointed these both to contribution/governance which gives an overview and starting point about our contribution process.


designertyler:
That didn't update the URL in the footer. How/where are the links defined for the footer?

Issue #1259, opened 2019-03-04
fix[spacing]: match spacing on intro content for guidelines pages to DL spacing


jeanservaas:
## Match intro spacing to DL spec

Intro paragraph and anchor links should be treated like one pattern; anchor links need to start much closer to the intro paragraph so content doesn't get pushed much further down the page. This pattern only occurs right now on the Guidelines pages... but we may bring intro paragraphs onto other pages eventually too.

- [x] Spacing between intro paragraph and anchor links should be 24px (right now it looks like 88-96px)

- [x] Spacing between anchor links and first H2 should be 80px (right now it looks like 88-96px)

<img width="1679" alt="image" src="https://user-images.githubusercontent.com/43144260/53769208-d1f9a880-3ea0-11e9-91af-db110aaed9f9.png">

jeanservaas:
@tw15egan sorry! Mike clarified it was 32px, not 24px, I was eye balling it. m'bad.

jeanservaas:
Hey guys, sorry to reopen, but we still need to make the above tweak... 

- [ ] it's 32px spacing after the intro paragraph (not 24px) to match IDL

![image](https://user-images.githubusercontent.com/43144260/54954341-6becc800-4f18-11e9-84fa-0620815ada45.png)

<br><br>

- [ ] also fix margin bottom spacing jump is happening between tablet and mobile (looks margin bottom goes from being 24px to 80px between those breakpoints)

<br>

![image](https://user-images.githubusercontent.com/43144260/54954373-7eff9800-4f18-11e9-92da-3303e9f0b5e8.png)

Issue #1258, opened 2019-03-04
bug[homepage cards]: Getting started cards are out of alignment / off grid


jeanservaas:
These cards should be flush with the right edge of the graphic like they used to be, they are correct in Safari but off alignment in Chrome and Firefox.

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/483104fd-469e-4bbb-be92-6d0554b8681f)

alisonjoseph:
this is a duplicate of #1252

Issue #1067, opened 2019-02-11
QA: Spacing Page fixes


jeanservaas:
Bug:
Can't get inline svgs to be the right color (even though it says they're right in the markdown (See spacing-02 and spacing-03 and layout-03)

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/1c6e30a4-38d3-4ba4-a5b7-f16e97c18e23)

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/8e1aa819-4af1-4563-900c-36026e5935eb)

Need to top-align in-line svg blocks with type (right now type is bottom aligned and svgs are top aligned in the fields.:

Enhancement:
![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/5d871b8d-ac2c-4b56-9ea1-dfbb3c23f94b)

jeanservaas:
Hey @alisonjoseph this was another fix I wanted to prioritize... at least to get those misaligned colors fixed... Mike thought we specifically changed the colors to emphasize certain spacers... he didn't realize it was a mistake. Just don't want to confuse people.

alisonjoseph:
👍 moved it to to do this sprint

Issue #1252, opened 2019-03-04
Homepage: Start designing & developing buttons are off grid


designertyler:
The images next to the buttons are stopping short of the grid. They should be aligned with the buttons.

![Screen Shot 2019-03-04 at 11.34.15 AM.png](https://images.zenhubusercontent.com/583dd918b9dc3c622022afbf/3619522d-4177-44b3-a8cf-031115aa626f)

alisonjoseph:
duplicate

PR #1262, opened 2019-03-04
fix(spacing): fix spacing issues around the site

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1258
Closes https://github.com/carbon-design-system/carbon-website/issues/1259
Closes https://github.com/carbon-design-system/carbon-website/issues/1067

- Fixed spacing on homepage tiles
- Fixed spacing on top anchor links 
- Fixed spacing images, aligned the rows to the top 

- Ensured all `h2`-`h4` are `display: inline-block;` so they properly respect margins placed on them


jeanservaas:
Much better, real quick:

- the distance between the intro paragraph and the anchor links needs to be 32px (not 24px) sorry, my bad, looks like it's still 24px

- Also, distance between anchor links and the first H2 (Carbon and accessibility) still looks too large by about 8px... can we just double check that it's 80px? 

- We miss you! But we know that you're following your heart by moving to Humbolt county to trim weed.

image

src/components/FeatureTile/feature-tile.scss
```diff
@@ -23,7 +23,7 @@ a.feature-tile {
   @include breakpoint('md') {
     width: 50%;
     position: absolute;
-    right: 0;
+    right: 1rem;
```
alisonjoseph: Hmm, your fix for #1258 here might conflict with my fix for #1252 here https://github.com/carbon-design-system/carbon-website/pull/1265/files#diff-63c095e1b95ae5b451ab9537ec133b54R35

I added the `ibm-col-bleed` class, but it seems @jeanservaas and @designertyler suggested two different ways to fix the misalignment. 

@jeanservaas @designertyler can you confirm if the image should bleed into the column, or if the card should be moved inside the grid.

alisonjoseph: I think the bleed is correct, based on how the other tiles are displaying, but I could be wrong. 
screen shot 2019-03-05 at 10 35 26 am

alisonjoseph: Or with the tile just moved in to line up inside the grid.
screen shot 2019-03-05 at 10 36 17 am

jeanservaas:
Hey guys, sorry I left this comment in #1259

This margin bottom distance is 24px now and i just need to change it to 32px! Thanks!

image

src/components/FeatureTile/feature-tile.scss
alisonjoseph: confirmed with @jeanservaas that your fix is correct 👍

alisonjoseph:
merging this in. will add spacing fix in another PR

shixiedesign:
@alisonjoseph Sorry I'm just looking after this for Jeanie so hopefully I'm not misunderstanding the goal here. Jeanie madea  comment that this spacing below should be 32px, or 2rem. It's currently 1rem it seems. 
image

alisonjoseph:
@tw15egan looks like one little update then we can merge this in

Issue #1261, opened 2019-03-04
bug[terminal code snippet]: show more button chevron is not right color 


jeanservaas:
## bug[terminal code snippet]: show more button chevron is not right color and it disappears

- [x] Chevron (icon) color should mirror color behavior of the "show more" text (use $icon-01 token)

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/35fb8bb3-3357-4e8e-bdcb-a6ba265d13a2)

jeanservaas:
@alisonjoseph for some reason I thought we already fixed this and merged it...  

![image](https://user-images.githubusercontent.com/43144260/54049977-ae39a980-41a3-11e9-948e-146e3d483145.png)

Ultimately this should be pulling from the Carbon component, just with different tokens, but there are still a lot of discrepancies

- [ ] there should be a gray 100 back ground behind show more (right now the code is visible under the button) 

- [ ] the code snippet icon is not the right one (it is correct in our component)

PR #1263, opened 2019-03-04
fix(icon): fix icon color

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1261

Changes chevron down color to `icon-03` (white)


jeanservaas:
Raaaal quick, just noticed this... and wanted to make sure that we're actually pulling from our code snippet component here:

- our Carbon code snippet component has the correct copy icon up in the right hand corner, but the one on the site is incorrect

- the show more chevron in the Carbon component is 16px (which is correct) and the chevron in the website component looks too large (it's probably 20px) 

Can we align with our component?

image

alisonjoseph:
merging this in, I'll fix the additional issues in the addon component I'm working on right now

Issue #1090, opened 2019-02-12
Pattern/Overflow page needs new animation & codepen


shixiedesign:
Pattern/Overflow page still contain old carbon design example images. 
https://next.carbondesignsystem.com/patterns/overflow-content

Animation at bottom of page: 
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/6576cd3f-0e69-461b-9f3d-cf750cec5454)

Code pen
![image.png](https://images.zenhubusercontent.com/5be09c3a6102900c6d16d991/ed7f2460-0943-457a-b4bd-3b21744b111f)

PR #1281, opened 2019-03-11
fix: overflow example animation

designertyler:
Closes https://github.com/carbon-design-system/carbon-website/issues/1090

Updated example animation to show the latest code example


aagonzales:
Should you maybe be showing the actual code-snippet component as the example instead of the website one which is a little different?

Also some of the #### are messed up on this page. Might be nice to also fix with this pr. 
image

Issue #1094, opened 2019-02-12
Enhancement[Icon libarary component]: Icon cards & icon search/filter bar


shixiedesign:
Sketchfile: https://ibm.box.com/s/rbrjv9hn9n8nia85z5f69tbt2ijazg2t
Video prototype: https://ibm.box.com/s/817f0vklb47nt4sg5f2mgwnchf8tkruy

YML file of data metadata: https://github.com/IBM/carbon-elements/blob/2d6bdaf676494e10e42f7ae0c8377b25d96330ec/packages/icons/metadata.yml

# Final icon/library page

### Default view of library with icon hover state, and dropdowns expanded
![Icon_Library](https://user-images.githubusercontent.com/15144993/54225151-100e5200-44c9-11e9-85c5-5ed7c11ef430.png)

### icon selected state
![Icon_Library expanded](https://user-images.githubusercontent.com/15144993/54225138-08e74400-44c9-11e9-81d3-798f678f5d9d.png)

### filtered state 
![icon_library filtered](https://user-images.githubusercontent.com/15144993/53374552-46bb6880-391e-11e9-95bc-fc473e8aa6a9.png)

### empty state
![icon_library empty state](https://user-images.githubusercontent.com/15144993/53374558-49b65900-391e-11e9-93fd-ec41e8cf0e46.png)

### mobile
![image](https://user-images.githubusercontent.com/15144993/54226076-04238f80-44cb-11e9-8827-dc31b826977d.png)

# Responsiveness

![image](https://user-images.githubusercontent.com/15144993/53374197-4e2e4200-391d-11e9-938d-3dc579a9b36b.png)

# Spec

![Icon page components](https://user-images.githubusercontent.com/15144993/54226018-e6eec100-44ca-11e9-8a4e-3860f4184477.png)

shixiedesign:
**Will add empty state design soon**

![image](https://user-images.githubusercontent.com/15144993/52662130-2f658f80-2ec9-11e9-96d4-a96cb92a71cf.png)

shixiedesign:
Empty state! 
![icon_library view 2 copy](https://user-images.githubusercontent.com/15144993/52664284-40fd6600-2ece-11e9-8dd3-20d24c9dcb05.png)

mjabbink:
@shixiedesign @jeanservaas 
I think the spacing between Categories is too much. 80 seems excessive. The Dl site seems good or it can be slightly more open maybe.

[...]

PR #1284, opened 2019-03-11
Add-disclaimer-icons

claycrenshaw:
Adds clarifying language to the top of the Icons Library page to prevent situations like this:

https://ibm-studios.slack.com/archives/GCPF37GEB/p1552319536014500

### DIFF
--- a/src/content/guidelines/iconography/library.mdx
+++ b/src/content/guidelines/iconography/library.mdx
@@ -4,6 +4,12 @@ title: Iconography
 tabs: ['Library', 'Usage', 'Contribution']
 ---
 
+## About the icon library
+
+This library page is for visual reference only. Please do not copy icons directly from this page for production work. 
+
+For designers, the icons are accessible as a Sketch library via the [IBM Design Language kit](https://www.ibm.com/standards/web/design-kit/). For developers, the icons are available via the [Carbon icons package](https://github.com/IBM/carbon-elements/tree/master/packages).
+
 import IconLibrary from '../../../../src/components/IconLibrary';
 
 <IconLibrary />


designertyler:
@shixiedesign has been working on the icon search page and had some sketches that use resources tiles for the links.

shixiedesign:
I did this the other day during our meeting. Two versions, either search/filter bar above the resource cards or after. I'd prefer the resource cards over the paragraph. The warning disclaimer is a very discouraging paragraph. I don' think it's necessary because 1 person decided to hack the site. 

If we really can't get agreement on a solution for this moment, hiding the library page until we've got a better experience is probably best. 

image

jeanservaas:
Quick thought on this... why don't we put the Usage tab first in the line up and include the Resource cards there.

We're breaking our Intro pattern on this page and there doesn't seem to be a good reason for it... also, it's a terrible experience to put the resource cards above the search or in between the search and the library.

image

I have the Resource cards (updated for all the elements) up in a PR I will link it to this issue. I did not move the Usage tab yet because I wanted to get people's thoughts @chrisconnors-ibm ?

chrisconnors-ibm:
**Please do not merge this PR.**

This issue is being addressed in https://github.com/carbon-design-system/carbon-website/issues/1094

mjabbink:
@jeanservaas Could not agree more to remove cards on top of Library. Those are for usage page.

mjabbink:
@jeanservaas @chrisconnors-ibm @shixiedesign 
The library could also have the cards at bottom of page after search and filtering results. As well as empty results, so at least you’ll get the resource cards rather than nothing.

mjabbink:
@chrisconnors-ibm @jeanservaas @shixiedesign 
Like this...

53374552-46bb6880-391e-11e9-95bc-fc473e8aa6a9

mjabbink:
Yes of course on top of usage. I specifically was talking about the library page

PR #1298, opened 2019-03-14
feat:Icon library resource cards added

designertyler:
A carbon user can access the icon assets from the icon library page so they can start using v10 icons in their designs. 

This temporary experience for v10 release adds two resource cards to the icon library until the icon card experience defined here https://github.com/carbon-design-system/carbon-website/issues/1094 can be built. The first resource card links the user to the icon repo, the second links to the IBM design language Sketch cloud.

Screen Shot 2019-03-14 at 10 40 54 AM


shixiedesign:
Maybe the copy can say "Get all icons via IBM Design Language Kit"

aagonzales:
I agree with Clay & Shixie, that was my one comment. It wasn't intuitive that the IDL kit is where the icons were.

Issue #1300, opened 2019-03-14
[SpacingPage] code example still uses old tokens


shixiedesign:
Spacing page's code example is still showing old tokens. This need to be updated to new spacing tokens `$spacing-01`, `$spacing-02`... etc
![image](https://user-images.githubusercontent.com/15144993/54393559-7b4a5680-4678-11e9-8731-fa21ae2fd688.png)

PR #1302, opened 2019-03-15
fix:spacing tokens

designertyler:
Closes #https://github.com/carbon-design-system/carbon-website/issues/1300

Updated the tokens in the spacing example
## Old (broken)
image

## New (fixed)
Screen Shot 2019-03-15 at 9 23 31 AM


shixiedesign:
I posted a question on `$spacing-auto` in https://ibm-studios.slack.com/archives/GCPF37GEB/p1552666045016800 . It might not exist coz i can't find it in both `carbon-component` and `carbon-elements`. When we hear back, if it doesn't exist, we can just remove it, and merge this PR!

PR #1313, opened 2019-03-18
UI shell docs revisions

claycrenshaw:
Major Overhaul: UI Shell Page

Added purpose section
Edits for accuracy & consistency
IA clarifications
Swapped Usage and Code tabs


claycrenshaw:
To Do:

- [ ] Update all images as needed to reflect current design

jeanservaas:
Hey @alisonjoseph — I was about to approve this PR, everything looks good to me but I'm wondering why the shell display is not so much longer than it used to be in the component display window... just to display more Latin text. The example used to be much shorter.

It's already live like this and doesn't technically affect this issue... so we can still merge, I was just wondering how easy it would be to fix. If we don't fix it now I will have to make a separate issue for backlog.

image

alisonjoseph:
With the update to v10 the example is coming directly from `carbon-components` so the example we had before was likely outdated. So as updates are made to the actual shell component we will see them here.

alisonjoseph:
We will need a separate UI Shell issue to look at the demo html.

PR #1317, opened 2019-03-20
Dl homepage video

rmoylan:
#6 - https://github.com/IBM-Design/design-language-website/issues/6

**New**

- Create Homepage video component ported from Design Language

**Changed**
**Removed**


vpicone:
Looking good! 

- I'm wondering if the `PlayPauseButton` and `VideoInternal` need their own component folder and story since they're only meant to be used in the context of the HomepageVideo.

- On that note, could the HomepageVideo just be `Video`?

- I'm, a fan of shedding the `Website` prefix for the components. But we should probably open another issue to rename old components.

alisonjoseph:
Agree 100% on naming, this library needs some cleanup/naming love so we can move it out of "alpha" 😉 

I'm not sure on naming it just "Video", since the styling is so different than just a standard video playing on the page. 🤔 I wonder if FeatureVideo might be better? you ok with leaving it for now and addressing it when we start to rebuild the IDL site @vpicone ?

vpicone:
Totally!

rmoylan:
@vpicone The InternalVideo will be used in multiple locations once the Design Language site is ported over to the Design Language site ( ex - https://www.ibm.com/design/language/elements/typeface/#our-typeface ). I made the HomepageVideo component its own component, as I didn't think there was going to be a way to place the overlaying links using markdown.

PR #1324, opened 2019-03-21
fix: use new carbon colors namespace

vpicone:
carbon-components no longer exports `ibm-colors__*` tokens. The name space switched to `ibm-color__*` and now the preferred usage is `carbon--*`.

In order to upgrade carbon-components, we need to use the newly exported namespaces.


alisonjoseph:
can you fix merge conflict when you get a chance 😸

PR #1329, opened 2019-03-22
fix(color pg): updated tokens

aagonzales:
Added:
- `$ui-background`

Updated role:
- `$interactive-01`
- `$interactive-03`
- `$interactive-04`
- `$link-01`

Updated color value
- `$support-02`


shixiedesign:
Not sure if this is an error, `Interactive-03` and `interactive-04` have the same colors in light themes and only 1 grade difference in dark themes, but one role say "4.5:1 AAA contrast", the other is "3:1 AAA contrast".

image

`Interactive-04` actually passes on ui-background 4.5:1 

image

PR #1330, opened 2019-03-22
fix(content): update v10 developer getting started and migration

### LINKED ISSUE #1305: Getting Started: Developers v10 content
Related to https://github.com/carbon-design-system/carbon-website/issues/1296, we need to make sure https://next.carbondesignsystem.com/getting-started/developers pages for Vanilla and React are up-to-date with v10 instructions and code snippets.

mattrosno:
Closes #1305

I made sure developer getting started content for vanilla and react is accurate with upcoming v10 release.

I also updated migration statement as carbon-icons won't be deprecated in v10.


vpicone:
* Noticed for `Other Frameworks` some of the headings aren't getting properly set. Maybe we could take care of that as part of this?

Screen Shot 2019-03-25 at 10 52 59 AM

* Should the resources go _under_ the anchor tags?
Screen Shot 2019-03-25 at 10 54 09 AM

 They look like they're running up against each other and it makes the navigation move around when switching to tabs without resources.

mattrosno:
I will address @vpicone's feedback now.

vpicone:
Nice, LGTM!

src/content/getting-started/developers/other-frameworks.mdx
```diff
@@ -3,15 +3,31 @@ title: Developers
...
+- [Troubleshooting](#troubleshooting)
+
+</AnchorLinks>
+
 The Carbon Design System supports vanilla JS, React and Angular as core parts of the product. But you can still build components even if you're using a different framework.
 
-**Using just the styles**
+### Using just the styles
 
 Developers wanting to build in different ways follow the instructions for the [Vanilla](/getting-started/developers/vanilla) library to access the styles and build out their own components.
```
chrisconnors-ibm: is this a behavior we want to encourage? "just using the styles" is a frequently encountered practice that is seem as equivalent to "using Carbon"; unless they would pick up future enhancements, we should be clear when this is OK and when it's Not OK. I assume this is really directed at teams creating their own frameworks of components rather than extending existing frameworks via submission or add-ons.

Issue #1141, opened 2019-02-18
[Website] Style tab QA


aagonzales:
QA content on all style tab component pages:
-  check for correct [color tokens](https://ibm.box.com/s/3prwmp421wu6umgayvejehiy0r1hi9m4)
-  check for correct [type token](https://ibm.box.com/s/3prwmp421wu6umgayvejehiy0r1hi9m4)s and values
-  check for correct [spacing tokens](https://next.carbondesignsystem.com/guidelines/spacing#spacing-scale) and values
-  change` SCSS` labels in table to appropriate token table (ie `Color token` or `Spacing token`)
-  when color section is too long split up into variation sections and/or "interactive states" section.
- change all dashes to en dash

aagonzales:
- [ ] Accordion
- [ ] Breadcrumb
- [ ] Button
- [ ] Checkbox
- [ ] Code Snippet
- [ ] Content Switcher
- [ ] Date Picker
- [ ] Dropdown
- [ ] File Uploader
- [ ] Form
- [ ] Link
- [ ] List
- [ ] Loading
- [ ] Modal
- [ ] Notification
- [ ] Number Input
- [ ] Overflow Menu
- [ ] Progress Indicator
- [ ] Radiobox
- [ ] Search
- [ ] Select
- [ ] Slider
- [ ] Structured List
- [ ] Tabs
- [ ] Tags
- [ ] Text Input
- [ ] Tile
- [ ] Toggle
- [ ] Tooltip

PR #1332, opened 2019-03-22
fix(data table): style tab tokens

aagonzales:
Related to: https://github.com/carbon-design-system/carbon-website/issues/1141

Updated data table tokens on style tab.


aagonzales:
Thanks @designertyler!

designertyler:
The images will be broken on the style tab since there are missing updated images. The new style images will be added in another PR.

Issue #1278, opened 2019-03-09
Design: Add Migration callout to homepage


claycrenshaw:
For launch, we need a prominent way to draw attention to the migration guide on the Carbon homepage.

Some options:

- Move Migration Guide link to top level in left nav (just after Getting Started)
- Upon GA, replace blue notification banner text (top of page) with a link to Migration Guide
- Add a visual tout of some sort to the "hero" space above the fold.

@jeanservaas @designertyler

chrisconnors-ibm:
the scope of this proposal exceeds (by quite a bit) the two days left before we are meant to be in code freeze. I would be willing to entertain:

> replace blue notification banner text (top of page) with a link to Migration Guide

…but don't know that @alisonjoseph has any time allocated to anything relating to that banner beyond removing it (probably doesn't).

@jeanservaas @designertyler what do you think?

in broader terms why is this coming up only now?

claycrenshaw:
Hey Chris, so I was told by Robin that there is no code freeze for the website content. 

I could move the side nav link in MD, that's very quick and doesn't require dev work.
I know that changing the text on the blue banner is trivial in terms of time required.

I'm happy to take either of the text/nav approaches if Jeannie and Tyler don't want to do something more visual.

chrisconnors-ibm:
sorry I was unclear about this. Dev resources/time are not the blocker. It is unvetted by @designertyler, @jeanservaas, or myself, an i'm not willing to approve capricious changes to the navigation structure this close to our planned release date. 

I've agreed to 66% of the proposal (temporarily repurposing the blue banner, even though it hasn't been resourced or planned for, or vetted yet, and a callout section on the home page, even though no one has been tasked with creating the accompanying imagery found in those callouts - only guessing that's what's meant - and I'm not sure who would create that), against my better judgement from a process standpoint - this is astonishingly last minute - why would litigating the 3rd thing be a priority at this point?

claycrenshaw:
Sounds good.

Carbon v10 is live! View the migration guide for help with upgrading from v9.

Inline link on Getting Started is a good idea too. Will sync with Connor on that.

designertyler:
The lifecycle of the banner was hopefully going to be short until the news of v10 percolates, we've had our announcements and blogs, and users figured out how to get back to the v9 site as they wrap up v9 work.

I could see that happening after a few weeks and then we can remove that bright blue banner from the top of every page on the site. 

[...]

PR #1338, opened 2019-03-25
fix(content):Text in banner for v10 release

designertyler:
Closes #https://github.com/carbon-design-system/carbon-website/issues/1278

Updated top banner text for v10 launch. Reference v10 migration guide.


vpicone:
We can use a Gatsby Link component for this now since the link is internal.

alisonjoseph:
Looks good! Let's hold on merging until v10 is actually live 😉

connor-leech:
Thoughts on "View the migration guide upgrade from v9" instead of "View the migration guide for help with upgrading from v9"?

Shorter and more action-oriented.

PR #1351, opened 2019-03-25
feat: add link to load special characters from google fonts

### LINKED ISSUE #1321: Arrows not plex on Safari
Arrows of the index links are not plex again. This happens in Safari, both desktop and mobile.

vpicone:
Closes #1321

**background:** 
- Safari no longer loads local font files other than system fonts. The google 
- fonts default stylesheet doesn't include the arrow character by default. 

**fix:** We can use the `text` beta feature of google fonts to specify the character and request that font file.

**new:** Add link to GF stylesheet with text query specifying our character. 

**note:** This feature isn't supported yet by the gatsby google fonts loading plugin. We can add encoded special characters to this link should we need more in the future.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -29,6 +29,8 @@ export default class HTML extends React.Component {
             rel="stylesheet"
             href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css"
           />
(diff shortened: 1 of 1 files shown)


vpicone:
@tw15egan It's a beta feature. What's wild is it doesn't even have to be part of the default subsets. As long as it's in the complete font file.

PR #1354, opened 2019-03-25
chore: update build step to remove storybook

alisonjoseph:
Netlify is timing out, removing the storybook build for now.


vpicone:
I think running them in sequence might be the next move once we get all these builds out haha

alisonjoseph:
this actually isn't needed

PR #1401, opened 2019-03-28
comment out toolbar for datatable until done

alisonjoseph:


designertyler:
@jnm2377 We don't need the pagination and small examples right now either, right?

jnm2377:
Oh yeah we haven't added the small styles so we should comment that one out too. Pagination data table is just pulling in styles from pagination so that one's ok to include.

jnm2377:
If we wanted to hide the actual toolbar in the data tables though, is that something that we have to do in the carbon-components repo or could we remove it from the html here? I can open up a PR if need be.

alisonjoseph:
I commented out the small one, @jnm2377 the data table is pulling html directly from carbon-components. Is there a version without the toolbar?

jnm2377:
ah, no, the configs have a toolbar from v9, I can submit a PR to remove the toolbar from the examples in that repo. 👍

PR #1403, opened 2019-03-28
feat: add automatic sitemap generation

vpicone:
Adds automatic sitemap generation. Sitemaps are useful for SEO and can also be used for screenshot generation: https://github.com/vpicone/sitemap-to-screenshots

#### Changelog

**New**

- `gatsby-plugin-sitemap`


gatsby-config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   siteMetadata: {
     title: 'Carbon Design System',
     titleInternal: 'IBM Product Design',
+    siteUrl: 'https://next.carbondesignsystem.com'
```
alisonjoseph: should we wait and update this to `www.` and merge it in once we're live? or do we need it now

PR #1405, opened 2019-03-28
Getting started images

### LINKED ISSUE #1336: Image Polish for Designer 'Getting started'
From @aagonzales on #1288:

> I think the content is good but the images could use a little help/polish. It would be nice to have a consistent screen crop or size throughout the page.
> Can we use the Carbon library in the library examples? Especially with symbols and the layer styles examples. I think we should be showing off the Carbon kit not the IDL one.
> If we do use the IDL kit can we make sure the image is properly calling the IDL kit. The example for layer styles is referencing the Template kit which 1. shouldn't be loaded as a kit and 2. its a place where you should be getting styles from
image

cc @connor-leech

jeanservaas:
Updated taco gif and static getting started images to clean them up a bit and add clarity.

closes #1336


jeanservaas:
closes this issue
https://github.com/carbon-design-system/carbon-website/issues/1336

PR #1408, opened 2019-03-28
fix: heading styles inside component docs

alisonjoseph:
Headings are all set to inline-block, overrides this for headings in a row on component doc pages so they have correct spacing.


jeanservaas:
Hey! sorry for the delay

There's definitely separation right now but a little too much, we just need 24px (looks like it's 48px now)

<br>

Component images:
image

<br>

rest of the site has only 24px:

image

alisonjoseph:
updated 👍 

Screen Shot 2019-03-29 at 10 39 29 AM

Issue #1061, opened 2019-02-11
Mobile: opening nav coz page to jump to top


shixiedesign:
On mobile, tapping on side nav jumps the page's scroll position back to the very top. 

![mobile-bg-jump](https://user-images.githubusercontent.com/15144993/52543714-590ba300-2d71-11e9-9703-abf469509f4d.gif)

PR #1453, opened 2019-04-04
sticky page tabs, and hide-able website header

lovemecomputer:
addresses the sticky behavior in https://github.com/carbon-design-system/carbon-website/issues/578 . i will create a separate issue to track implementation of the tab-stacking behavior described within

**todo:**
- [x] implement 3px top border on page tabs for a11y reasons
- [x] on scroll down, add class `.website-header-hidden` to `<body>` (or any other high-level parent)
- [x] on scroll back up, remove the class
- [x] add some 'throttling' so the header doesn't hide/show until you've scrolled a small amount
- [x] website header: test animation/transition performance, and consult https://www.sysleaf.com/js-toggle-header-on-scroll/ et al for improvements
- [x] ~for some reason  opening the side nav menu jumps the scoll to the top—verify if this is happening in current site or if it is a bug introduced in this PR~ : known issue tracked in https://github.com/carbon-design-system/carbon-website/issues/1061

## states and elements to verify when testing:
- element: sticky global header
  - states:
    - hidden
    - visible
- element: sticky page header
  - states:
    - hidden
    - visible
- element: side nav
  - states:
    - hidden
    - visible
- element: site switcher
  - states:
    - hidden
    - visible
- element: website alert
  - states:
    - present
    - not present
(website alert pushes everything's sticky state down by 2.5rem)
- element: page tabs
  - states:
    - page with no tabs
    - page with tabs

2^6 = 64 separate states the site can be in. i tested this across chrome, safari, and firefox, but would appreciate a couple extra eyes 😇


src/components/WebsiteAlert/website-alert.scss
```diff
@@ -0,0 +1,178 @@
...
+.website-alert__text {
+  @include carbon--type-style('body-short-01');
+  color: $inverse-01;
+  padding: 0;
+}
+
+.website-alert__text span {
+  &:nth-of-type(1),
+  &:nth-of-type(2) {
+    font-weight: 600;
```
alisonjoseph: We should update any new code (and old when we have time) to use

`font-weight: carbon--font-weight('semibold');`

src/components/PageTabs/page-tabs.scss
```diff
@@ -4,6 +4,7 @@
   height: rem(48px);
   position: relative;
   border-top: 1px solid rgba($ui-02, 0.25);
+  // background: $carbon--black-100;
```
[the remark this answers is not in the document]
lovemecomputer: will do! i was testing out different nestings but think i settled back on something that doesn't need this rule

src/components/WebsiteAlert/website-alert.scss
lovemecomputer: ah thanks!! i just copy-pasted without looking. will look over & update!

lovemecomputer:
looked into making the state switching & class name propagation into a declarative implementation, but it was causing full-page re-renders, or bumped up against other architectural issues. 

will implement this as-is, with `this.stickyWrapper = document.getElementsByTagName('body')[0]` & `      this.stickyWrapper.classList.remove(this.headerHiddenClass)` etc to get the desired behavior on our site state.

in the future i plan to pair with @vpicone on the gatsby theme to port over the functionality into a nice declarative way!

all that's left now is some polishing of the interaction, and transition performance

src/styles/_sticky.scss
```diff
@@ -0,0 +1,148 @@
...
+  .website-alert
+    + .bx--header
+    ~ .bx--side-nav
+    + div
+    .color-token-table__theme-switcher--sticky {
+    top: 104px;
+  }
+}
+
+// .website-alert + .#{$prefix}--header ~ .#{$prefix}--side-nav {
```
alisonjoseph: can we remove unused code?

alisonjoseph:
+1 to what @laurenmrice said about pages without tabs. What do we want to happen here from a design perspective?

lovemecomputer:
@laurenmrice thank you, i totally missed this!! 

@alisonjoseph atm this is not intended as i understood it, this is a bug because i only tested locally on pages with tabs and didn't see this behavior. it's happening because the header which contains the tabs is fixed at the same height whether or not it contains tabs. so i'll need to think about a way to make it a different height when it doesn't have a child of `PageTabs` ig.

laurenmrice:
For pages with no tabs, We would just want the top header to remain sticky, and for the rest of the content on that page to scroll. Example of what i mean on the IDL site. : https://www.ibm.com/design/language/elements/type-basics

jeanservaas:
omg audrey this looks really good!!!!

Aside from Lauren's point I'm ready to approve.

lovemecomputer:
@laurenmrice @jeanservaas: re: alison's comment, this PR currently includes all fixes pointed out above, in this thread, aside from the global site header being visible on pages without tabs. as she said, unfortunately that involves a bit of a change to how we are architecturally passing around data about the content of the current page, and it would be best addressed after we merge in the big re-architecting of site packages/themes that vince has been leading.

(i also personally think it's kinda weird/inconsistent behavior to have it the global header to be sticky on some pages and not others. but this can be a separate conversation in a separate issue/PR for implementing the behavior 😇)

tw15egan:
🎉

Issue #1470, opened 2019-04-08
bug fix[Grid]: homepage and resources page


jeanservaas:
## bug fix[Grid]: homepage and resources page

<br>

After PR #1198 was merged the latest articles on the homepage and Featured and Articles on the Resources page do not have the right grid designation... @shixiedesign tried to change in markdown but could not figure out how.

- [x] Use "no gutter on the left side" grid designation for these article tiles; this allows article tile to hang only into the left hand gutter so text will align. There should be 16px gutters between the tiles (not 32px) 

<br>

*Correct spec:*

Note how resource tiles hang into the gutters on both sides, whereas article tiles hang into gutter ONLY on the left side.

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/ebeb62f5-46a4-4be0-9e91-de837f9cb161)

<br>

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/4fdf3127-6c03-48e5-bf57-b4896824fc8e)

shixiedesign:
I made a couple attempts at changing the prop from `noGutterMdLeft` to `noGutterLeft`, `noGutterLgLeft` etc, but it doesn't seem to be affecting the layout. 

```
<Column offsetLg="4" colLg="4" colMd="4"	noGutterLg>
```

PR #1471, opened 2019-04-08
chore: update breakpoint include prefix

alisonjoseph:
Closes #https://github.com/carbon-design-system/carbon-website/issues/1470

Grid component css was missing updated breakpoint mixin `carbon--`

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.0.1-alpha.19",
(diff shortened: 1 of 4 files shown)


shixiedesign:
I'm trying to review, assuming this PR affects breakpoint style changes. Is this an issue due to this PR? See **Developer getting started** page has 5 tabs, and the 5th one cannot be accessed at `sm` breakpoint because the tabs didn't turn into a drop down. 
image

PR #1480, opened 2019-04-10
feat(homepage): add homepage video component

### LINKED ISSUE #1273: update[Home page] replace still image with video loop and add pause button to lower right corner
## update[Home page] replace still image with video loop 

- [ ] add pause button to lower right corner (per specs) for accessibility (otherwise video loops)
- [ ] overlay is #000000 with 0.7 as the alpha over video
- [ ] make homepage dot pattern slightly brighter (@shixiedesign Mike has asked for this, whichever dev is assigned, you can see if we can get this done also)

Pause button design spec:

image

Attached visual issue:
https://github.com/carbon-design-system/carbon-website/issues/1089

Vimeo link:
https://vimeo.com/322021187/eb052e5feb

jnm2377:
Closes #1273 

Adds homepage video to header

#### Changelog

**New**

- `<HomepageVideo/>` component

**Changed**

- `<HomepageHeader/>` includes video instead of image
- change dots opacity from `0.25` to `0.35`

**To-Do**:
- [x] fix video wrapper height to be the same as parent height (not overflowing like currently)
- [x] change pause button positioning to bottom instead of top (need to fix video wrapper height first bc it's overflowing and therefore the video controls aren't showing if they're positioned at the bottom)


tw15egan:
This looks awesome!

jnm2377:
@tw15egan thanks!! It definitely makes a huge difference to the homepage! 😄 
@shixiedesign just pushed final updates, so it's ready for review. Also, the overlay opacity right now is `0.7`, and I changed the dots to `0.35` so they're brighter. Lmk if either of those need to be changed. 👍

jnm2377:
@shixiedesign I pushed some changes to try to center the video more. It's tricky, it's not actually receiving any sort of vertical alignment styles, but I added media queries for the top positioning to make the `iframe` more responsive at different widths since I think it was mostly cutting off the bottom at large widths since the ratio of the video and the header are different.

shixiedesign:
Hey looking good! Can we reduce overlay just a bit more, maybe `0.6`? Then we are good!

lovemecomputer:
good work @jnm2377 !!

one thought: the video can take a while to load, which means loading the site you see 1) black background, and then 2) the background video instantly appears. 

could it fade in from 0 opacity, or some other transition to smooth that out? (it's quite a long delay for me here on the IBM wifi, so it was pretty noticeable)

PR #1482, opened 2019-04-11
chore: bug fixes

### LINKED ISSUE #1031: QA [Component Display Code Snippet] -- offsite links missing hover state
Add-on code snippet component (the one attached to the component display window) is missing a hover state for the "React" "Angular" and "Code pen" links

This is m'bad. Did not specify hover state to @tw15egan — I didn't notice this because we don't have any of these links in the components right now, but it showed up in the UI Shell Component.

- add underline for hover state on React off site link (just under text, like Show more below)
- make sure React and icon are both blue 40 (there is no color change for hover)
- make sure "Show more" and arrow icon, "Code:" and "Vanilla JS" are all Gray 10 (they look white right now)

Now: no hover on React (it needs an underline on hover like "Show more")

### LINKED ISSUE #1371: Remove extra radio button examples
Screen Shot 2019-03-27 at 9 50 23 AM

This appears twice in the website and the title doesn't correspond the the example.

### LINKED ISSUE #1376: [QA bug] Links lost duo styling
Seems like website's links lost our custom styles. That default color is not Blue 60, and there shouldn't be a visited color. 
https://next.carbondesignsystem.com/getting-started/designers
image

### LINKED ISSUE #1397: Style fix for component display window's radio buttons
Radio button on component window:
- [x] Should be dark theme radio buttons
- [x] Should be right aligned so it doesn't get cropped. 

**Current:**
image

Spec for radio button in component display window is here:

**Should be:**

alisonjoseph:
#### Changelog

- remove visited link color Closes #1376
- update radio button colors on component pages Closes #1397
- add hover underline to external library links on component pages Closes #1031
- remove extra radio button example Closes #1371


alisonjoseph:
@shixiedesign good catch, updated to use the styles from carbon-components with the bx--link class name. (at 16px instead of 14px though)

src/components/markdown/Markdown.js
```diff
@@ -173,3 +173,11 @@ export class p extends React.Component {
     );
   }
 }
+
+export class a extends React.Component { 
+  render() {
+    return (
+      <a className="bx--link" href={this.props.href}>{this.props.children}</a>
```
vpicone: maybe spread the props and do the classname thing in case folks want to build on it

PR #1504, opened 2019-04-17
chore(dependencies): upgrade to MDX 1.0

asudoh:
This change upgrade MDX from `1.0-alpha` to `1.0`.

#### Changelog

**Changed**

- Dependencies to use MDX `1.0`.


alisonjoseph:
Looks like this is breaking any of our components on the site that allow for nested markdown inside them. (anchor links, tiles, tabs...)

asudoh:
@alisonjoseph Thank you for catching that - We needed another migration described in: https://github.com/ChristopherBiscardi/gatsby-mdx/issues/239#issuecomment-468484751

alisonjoseph:
Looks like we are missing the `.wrapper` div now, which is causing some spacing issues at the top of pages, and below paragraph tags (might be more, haven't dug in to it yet).

asudoh:
Just to leave a note wrt the talk @alisonjoseph and I had offline - @alisonjoseph kindly created asudoh#1 for wrapper `<div>` and I have manually merged the change to this PR. Thanks @alisonjoseph!

asudoh:
Got it - Thank you for catching this! Updated.

PR #1505, opened 2019-04-17
fix: Type scale style fixes

### LINKED ISSUE #769: fix[Type scale exhibit]: Not to design spec
## fix[Type scale exhibit]: Not to design spec

- [x] Need 16px top margin before "rem" and "px" start

- [x] "Px" column needs to be moved to the left (32px distance from end of "rem" column per IDL spec

- [x] All type in these exhibits should be Gray 100 ( # 171717)

- [x] Space out the scale and cap align the the rem with the example to the right. (i.e. more line height to match IDL example below) 

<br>

This type should also be Gray 100:

<br>

Cap alignment:
image

alisonjoseph:
Closes #769

Typography page component fixes

@jeanservaas if the alignment isn't 100% right, let's sit down together and tweak in person. It's built differently from IDL so unable to get it exactly the same.


alisonjoseph:
updated @jeanservaas

jeanservaas:
@alisonjoseph  thanks for quick work on this, looking good! Couple things:

1) The drop shadow only shows up on the bottom of the breakpoint changer component when it's sticky and content is under it. 

So it should look like this initially, before the user starts scrolling and it becomes sticky:

image

This is what it looks like now:
image

<br>

2) We should be using the same drop shadow as the color usage component and the drop shadow only appears beneath the component, not all around it -- like this 

image

<br>

Also could we add a little vertical height to the white containers... on Carbon they are all 176px and on IDL they are all 216px (even the body styles etc.) We'd also have to update this on the productive page too. Basically, until the spec type wraps below at smaller breakpoints, the white containers all stay at 216px vertical height.

image

alisonjoseph:
@jeanservaas best to make another issue for the do-dont aspect ratio issue so it doesn't get lost.

alisonjoseph:
@jeanservaas ready for re-review.

jeanservaas:
1) Almost there, really sorry this is so fussy, they're not using spacers from our scale to get some of these line heights on the type scale exhibit. Your preview link basically looks good with the line height, the only place where it looks a little weird is the area between .75rem - 1.25rem

The exact margin-top for those on IDL are:

.875 rem (margin top = .315rem)
1 rem (margin top = .36 rem)
1.125 rem (margin top = .405 rem)
1.25 rem (margin top = .45 rem)
1.5 rem (margin top = .54 rem)

If you want to get together on Monday and tweak in person we can also do that and then just merge it.

2) Don't kill me. All the "Plex XX"s (below .125) are riding a little low of the hang line; it's because the top of the 'l' ascender is what's hitting the hangline... but it should actually be the top of the "P" that aligns.

image

3) Can we remove "rem" from after the numbers, we have it in the top as a category, so seems redundant and will clean up design a little bit; I may not have noticed this before, so sorry about that.

image

We're almost there. please don't hate me.

alisonjoseph:
@jeanservaas maybe we can sit down together for 15 min tomorrow and figure out the spacing. The spacing is set as padding-bottom on the first column. 

Screen Shot 2019-04-22 at 11 08 58 AM

PR #1511, opened 2019-04-17
Fix:Update to carbon component version to 10.1.0

designertyler:
The data table component examples are not the latest. @jnm2377 suggest updating to the latest version of the components would be the fix.

Current data table example
Screen Shot 2019-04-17 at 4 46 04 PM


designertyler:
Closing. Will open another soon

PR #1513, opened 2019-04-18
chore: Update spacing tokens

alisonjoseph:
Updates to spacing tokens to use new tokens
Adds tokens across site to margin/padding

### Testing
- double check I didn't replace anything with the wrong tokens 😅  

_might be easier to review if you hide whitespace changes_


src/components/ComponentExample/component-example.scss
```diff
@@ -43,7 +43,7 @@
   bottom: 0;
   color: $text-02;
   @include carbon--type-style('caption-01');
-  padding: 0.5rem 1rem;
+  padding: $spacing-03 1rem;
```
jnm2377: did you want to update the `1rem` to also use a spacing token?

src/components/PageTable/page-table.scss
```diff
@@ -26,7 +26,7 @@
       td {
         @include carbon--type-style('body-short-01');
         color: $text-01;
-        padding: $spacing-md;
+        padding: $spacing-05;
 
         img[src$='checkmark--glyph.svg'] {
           width: 16px;
```
jnm2377: semi related to this PR. Would we want to also update layout tokens? If they're being used or not used, like here?

src/components/ComponentExample/component-example.scss
alisonjoseph: good catch 👍

src/components/PageTable/page-table.scss
alisonjoseph: We're not currently using layout tokens anywhere on the site I don't think, but would probably make sense to do that at some point.

PR #1525, opened 2019-04-19
fix:Add charts to FAQ

designertyler:
This adds the FAQ about where charts and data vis went in v10 and links to the in-progress charts work.


connor-leech:
I'd like the language to be a bit more firm, and we can't have non-contextual hyperlinks.

Also, just as a note while I'm here. We should decide if/where we're going to speak in first person. I'm not wholly opposed, but our content is pretty inconsistent on that front.

"Carbon charts are in development and classified as experimental. Check out the experimental charts repo to explore the work so far.

These charts will receive an updated color palette to match Carbon v10. Visual guidance and chart components are also in the works!"

Issue #1551, opened 2019-04-25
Figure out how to distribute 4 theme kits 


aagonzales:
There are now 4 separate Sketch files, one for each theme, because anymore more than two in one files caused Cloud upload problems. We'll need to figure out how we want distribute the 4 kits. Possible options:

- 4 separate resource tiles (maintain the one click to download)
- 1 resource tile that links to the[ Sketch Cloud](https://sketch.cloud/documents/all-documents) page and from there the user chooses which theme library to download.
- Download a disk drive like Apple's flow
- Is there a way to one click bundle download multiple libraries at once?
- Some sort of Sketch Plugin? 

We can also have an immediate solution and then create a better long term (wedding cake) solution. 
- [x] Cupcake solution
- [ ] Wedding cake solution

aagonzales:
Cupcake: has been added as 4 separate tiles on the resource page and the designer/getting started page.

Closing for now as other issues are higher priority and this is satisfactory for now. Other solution require Sketch to make update or us to build something.

PR #1556, opened 2019-04-26
Kit resources

aagonzales:
Now that there are 4 theme libraries we couldn't link to just one file/download anymore. Needed a way to link out to the 4 separate kits.

**Changed:**
- Add library tiles to the resource page
- Change "Design Kit" link in left nav to go to resource tab (mimics the same interaction as the ithub link)

**Related:** 
https://github.com/carbon-design-system/carbon-website/issues/1551


connor-leech:
Can we hold off on merging this until we settle on header language for Getting started? Want to be consistent across both pages.

PR #1559, opened 2019-04-26
[docs] Add themes guidance to Designers/getting-started

connor-leech:
Related to #1551

This isn't our most elegant solution, but the association between resources and steps is more clear in-line. Will look better with baby tiles and, hopefully, an even more streamlined Sketch solution 🎉 .


connor-leech:
Please do not merge this PR. We will be merging in a coordinated effort on Monday.

PR #1567, opened 2019-04-29
fix: sticky nav on color page

alisonjoseph:
Noticed the math was off on the sticky tabs on the Color page after merging in the sticky page level tabs. 

Fixes when/where color page tabs become sticky


lovemecomputer:
good catch!! there is one small issue still:

Apr-29-2019 15-44-06

alisonjoseph:
@lovemecomputer it should be fixed now. still not 100% when scrolling fast, but it should be in the right position now

lovemecomputer:
okay cool thanks!! i’m seeing a little issue on mobile which is: after scrolling up to reveal the sticky header, it obscured the page tabs. happens on this page 


4DDADAC5-5550-432A-B9CD-240F32C4EE20
D31C3446-BE47-4F35-A2C5-849C8ECB1E8C

however, i see a couple of other mobile issues, so maybe that would be best handled in a separate issue & PR for mobile fixes 
- color page “White” theme label is truncated to “Wte”?
- transitions for sticky header not happening in iOS Chrome—everything just snaps into place

alisonjoseph:
I think "Wte" was on purpose? 🤷‍♀ but yea, we might want to address these other issues in the theme, so we don't duplicate work. 😅

tw15egan:
Noticed the nav does not match the width of the table below it when it sticks, seems to be about 2-3px short


Not sticky:
Screen Shot 2019-05-01 at 10 43 33 AM

Sticky: 
Screen Shot 2019-05-01 at 10 43 38 AM

tw15egan:
Guess it's just 1 breakpoint: 

Screen Shot 2019-05-01 at 10 50 25 AM

alisonjoseph:
@tw15egan what breakpoint are you seeing that? I'm not able to duplicate it 🙈

tw15egan:
I think it's only between `1280px` and `max`, so not a big deal at all 😄

alisonjoseph:
@tw15egan ah ha! I see it. fixed.

tw15egan:
Woo! 🎉

PR #1580, opened 2019-04-30
fix: sending resource tile to correct location

connor-leech:
Sketch library resource tile on homepage was sending users to White theme. Now directs to Themes location on Resources page.


alisonjoseph:
Approved this for now, however, tiles weren't built for internal links, so this is linking out and opening a new window instead of linking to the Resources page. It also has the external link icon vs. the arrow icon.

Issue #1558, opened 2019-04-26
Code snippet copy button not working on some pages/components


lovemecomputer:
copy button for code snippet does *NOT* work on the code snippets on:
- https://www.carbondesignsystem.com/getting-started/developers/vanilla

it *does* work on these kinds of code blocks:
- https://www.carbondesignsystem.com/components/link/code

tw15egan:
We don't ship any copy to clipboard functionality out of the box, so we probably just need to wrap these copy buttons like https://github.com/carbon-design-system/carbon-website/blob/ae1f9ba29d55fe85ebe7aa762229352e855f5c79/src/components/CodeExample/CodeExample.js#L85

PR #1595, opened 2019-05-02
fix(copy-button): provide text to copy button

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1558

Adds in the ability to actually copy the text on pages like https://www.carbondesignsystem.com/getting-started/developers/vanilla

Multi-line commands still cause problems when copied directly into the terminal. Wonder if we could possibly replace the `$` with a new line character?

#### Changelog

**Changed**

- Provided `CopyToClipboard` with a `text` prop


alisonjoseph:
I'd be ok with replacing the `$` if we need to for multi-line snippets

tw15egan:
I think I need to clean this up a bit... It is removing `$` from `scss` variables. I'm going to take a look and see if I can only remove ones that have a space following.

tw15egan:
@alisonjoseph alright, got it matching just `$ ` and not `$variable`

PR #1641, opened 2019-05-17
docs: adding or creating pattern

arnaudgillard:
{Arnaud Gillard submitting patterns from the Cloud Integration Design Guidelines}

#### Changelog

**New**

- {Complete new page named "Adding or Creating" living under the Patterns section}

**Changed**

- {All New}

**Removed**

- {All New }

PR #1655, opened 2019-05-22
 docs: Creation flows

arnaudgillard:
{Arnaud Gillard submitting patterns from the Cloud Integration Design Guidelines: Submitted branch pull request as requested in #1641}

Changelog

New

{Updated the "Adding or Creating" living under the Patterns section. Renamed the page to "creation flows" and re-structured all content.}
Changed

{All New}
Removed

{All New }


src/content/patterns/creation-flows/creation-flows.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+- [Standard creation](#standard-creation)
+- [Multistep creation](#multistep-creation)
+- [confirmation](#confirmation)
+
+</AnchorLinks>
+
+
+## Standard Creation
+
```
aagonzales: Maybe add an intro sentence or two about what makes up or qualifies something to use a standard creation pattern. Something that makes the user understand/chose why a standard flow might be the better choice than the Multi-step flow.

src/content/patterns/creation-flows/creation-flows.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+<ImageComponent cols="8">
+
+![Example](images/3.png)
+
+</ImageComponent>
+
+### Full page
+Consider using this when wanting to take full advantage of the screen’s available real estate.
```
aagonzales: Maybe elaborate on the why you would want to take advantage of the full page. Like "best for when you have a longer creation process that involves multiple form items to be completed" or something a long those line. You hit on the why in the modal and side panel but not so much this one. The examples shows it but it would be nice to actually say it.

src/content/patterns/creation-flows/creation-flows.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+
+<AnchorLinks small>
+
+- [Side panel](#side-panel)
+- [Full page](#full-page)
+
+</AnchorLinks>
+
+### Side panel
```
aagonzales: Here I think you can maybe repeat the benefits of having it in a side panel like in the "standard creation section" incase the user doesn't read that.

src/content/patterns/creation-flows/creation-flows.mdx
```diff
@@ -0,0 +1,164 @@
...
+<ImageComponent cols="8">
+
+![Example](images/10.png)
+
+</ImageComponent>
+
+### Full page
+<ImageComponent cols="8">
+
+![Example](images/11.png)
```
aagonzales: I think maybe break this image up into 3 separate images so we can caption specifically what's happening in each to help the user understand the differences.

arnaudgillard:
Hey Anna! Sounds great. Thanks for your feedback! Yes I certainly agree that we should try and get this as polished as possible. In fact, as you brought this up, I checked with a visual and there are indeed elements which need checking. I was sent these files by the visual team so will check with them why it's not accurately following the carbon guidelines. Will push an update once I get them refined.

src/content/patterns/creation-flows/creation-flows.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+When any create action is a success, a notification banner can be presented to visually confirm this. 
+
+<ImageComponent caption="Example of a successful creation confirmation" cols="8">
+
+![Example of a successful creation confirmation"](images/12.png)
+
+</ImageComponent>
+
+#### Maintainers:
```
alisonjoseph: Can we move the maintainer's section to the top of the page, afraid it will be missed at the bottom.

arnaudgillard: Updated!

arnaudgillard:
Hi all, I've unfortunately managed to push an update on a separate commit (although I tried staying within the same branch!). Here is the link to the updated page: https://github.com/carbon-design-system/carbon-website/pull/1689

arnaudgillard:
Just adding a comment here to avoid it from closing :)

src/content/experimental/creation-flows/creation-flows.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+</ImageComponent>
+
+### Full page
+
+If your creation flow consists of multiple fields and/or supporting imagery, consider the full page option. Full page creation flows allow you to take advantage of maximum screen real estate.
+
+<ImageComponent caption="Example of a full page creation flow" cols="8">
+
+![Example of a fill page creation flow](images/8.png)
```
aagonzales: I think I would arrange this page more like this:
- Unless buttons are attached to the edge of a container (image doesn't count) then the buttons should have a 8px space between them. 
- They should also be aligned with the title at top. Base align the text of the title with the text of the button.
- Could you also fix the hanging e in "workspace"
image

arnaudgillard:
request from @jnm2377 :

Can we use the do-don't components for this set of example images? You can see an example here: https://www.carbondesignsystem.com/guidelines/color/overview https://github.com/carbon-design-system/carbon-website/blob/master/src/content/guidelines/color/overview.mdx

arnaudgillard:
please refer to changes made in #1796 whilst I try to get that content onto the correct branch (this one).

arnaudgillard:
Just copied the changes from #1796 over here so that its all in one pull request. I have updated the columns of the imagery and used the do/don't component from the carbon page. I will close #1796. I also have our visuals making the requested changes so will get that to you soon @aagonzales.

connor-leech:
This work has been transferred over to the new website repo: https://github.com/carbon-design-system/carbon-website/pull/121

PR #1688, opened 2019-05-31
docs: tooltip guidance/image

### LINKED ISSUE #1640: [Tooltip, website] Update tooltip page with new alignment options doc
Recently merged enhancements for tooltip. Now tooltip supports 3 alignment options: start, center, end (previously only center is available). This should be documented in Usage section: 
https://www.carbondesignsystem.com/components/tooltip/usage

Dev isseu: https://github.com/carbon-design-system/carbon/issues/2678
Design kit issue: https://github.com/carbon-design-system/carbon/issues/2768

### Position and alignment
Tooltip maybe positioned `top` or `bottom`, and aligned to `start`, `center` or `end`. Please connect with @laurenmrice for image assets.

shixiedesign:
Closes #1640

@connor-leech added requested guidance, I polished image and text a bit. 

All changes are in https://carbondesignsystem.com/components/tooltip/usage

#### Changelog

**New**

- Added image from @laurenmrice 
- Added corresponding sentence about placement of tooltips

**Changed**

- Definition tooltip usage image and copy


asudoh:
Hi @shixiedesign thank you for the update here! One question on interactive tooltip - I originally was imagined that interactive tooltip has very limited use case, but [this image](https://github.com/carbon-design-system/carbon-website/blob/f1e15d9fe8424183b1d21b613a0bf65189e0c3c7/src/content/components/tooltip/images/tooltip-usage-3-v2.png) seems to tell that there are more use cases - Is it correct? Wondering if "anything that contains interaction" goes to interactive tooltip, or there are some different usage definitions. Currently the API and implementation details of the interactive tooltip code has very limited use case in mind - Wondering if we need to change that. Thoughts will be nice - Thanks!

shixiedesign:
@asudoh I agree we should be using interactive tooltip for limited situations, hence the added sentence:

> Please note that in general hiding interactive content in tooltip is discouraged. 

Do you think the build/implementation of interactive tooltip doesn't allow it being used on product walkthroughs or on-boarding tours? If so, we should open an issue to address it – sorry because I realize this info wasn't in the original spec. Walkthroughs & on-boarding are the main use cases for this component, at least today. If you've seen it in other situations, let us know!

src/content/components/tooltip/usage.mdx
```diff
@@ -2,9 +2,35 @@
...
 ---
+
 ## General guidance
 
-Tooltips provide additional information upon hover or focus.
+Tooltips provide additional information upon hover or focus. The information should be contextual, useful, and nonessential information. Keep tooltips short.
+
+## Placement
+
+Icon tooltips and interactive tooltips may be positioned **top**, **below**, **left**, or **right** to the trigger item. The container of the tooltip text may be aligned to **start**, **center** or **end**.
```
aagonzales: Super knit-picky but should it say "bottom" instead of "below

connor-leech: I actually think that it should be "above" and "below" 😂, but pretty much agree

asudoh:
@shixiedesign Thank you for sharing your thoughts and my apologies for somehow missing your reply:

> Do you think the build/implementation of interactive tooltip doesn't allow it being used on product walkthroughs or on-boarding tours? If so, we should open an issue to address it – sorry because I realize this info wasn't in the original spec. Walkthroughs & on-boarding are the main use cases for this component, at least today. If you've seen it in other situations, let us know!

Yes it'll be a good idea to track it an issues. In this way we can create live examples and can ensure interactive tooltip works in those scenarios. The current tooltip code almost assumes limited scenarios wrt trigger link/buttons, so good to gather the use cases.

src/content/components/tooltip/usage.mdx
shixiedesign: Right, language wise I don't mind changing to any of your options. But to follow the code terminology, I think I will change it to "top" and "bottom" if that makes sense to everyone. 
image

PR #1696, opened 2019-06-03
docs: Add guidance to modal

### LINKED ISSUE #1695: Add guidance to modal
### 1. One & three button usage

If a modal only has one button, the button should take up 100% of the container. 
**Side-note**: modals that have no real action to take (like ok) should be using the passive modal if its just informing the user and not asking the user to take an action.

image

Modal should not have 3 buttons. The "help" or other 3 non-primary action should be added as a link in the content of the modal. 

image

_Originally posted by @aagonzales in https://github.com/carbon-design-system/carbon/issues/2325#issuecomment-484988200_

### 2. Elements in modal takes up 100% or 75% width

- Components (eg. data table, form, progress indicator) should take up 100% of modal width. 
- Text & paragraph component should take up 75% only

connor-leech:
Closes #1695

Added requested modal documentation.

#### Changelog

1. One & three button usage
2. Elements in modal take up 100% or 75% width
3. Cleaned up.

#### Notes

I elected not to include the image assets in the related issue. This is the perfect opportunity for a spec image, and we should create one. This is not stopship, we can merge and add later if nobody can get to the artwork in time.

Request [here](https://github.com/carbon-design-system/carbon-website/issues/1697).


src/content/components/modal/usage.mdx
```diff
@@ -3,32 +3,43 @@ title: Modal
...
 ---
 
+### Modals communicate information via a secondary window and allow the user to maintain the context of a particular task.
+
 <AnchorLinks>
 
 - [General guidance](#general-guidance)
 - [Format](#format) 
 - [Variations](#variations) 
-- [Dismissal](#dismissal)
```
shixiedesign: Looks like this is no longer a stand alone section, and a sub section of **General guidance**. H2 index link should be removed.

src/content/components/modal/usage.mdx
```diff
@@ -3,32 +3,43 @@ title: Modal
...
-_Modals_ communicate information via a secondary window and allow the user to maintain the context of a particular task. Use modals sparingly because they interrupt user workflow.
+Modals interrupt user workflow by design. They are most effective when a task must be completed before a user can continue. While effective when used correctly, modals should be used sparingly to limit disruption to a user experience.
+
+#### Dismissal
+
+Modals may be dismissed in 3 ways:
+
+- Using the “✕” in the upper right-hand corner of the modal
+- Pressing the `ESC` key
+- Clicking or touching outside of the modal
```
shixiedesign: I'd like to double check with @carbon-design-system/design on this. It makes sense that Passive modal can be dismissed by clicking outside the modal, Transactional modal should probably not be dismissed so easily, in case of accidental clicks outside before task completion. If everyone agrees, I can make an issue for dev fix.

src/content/components/modal/usage.mdx
```diff
@@ -3,32 +3,43 @@ title: Modal
...
 
 ## Format
 
+The modal is composed of three distinct zones: A header, the body, and a footer. Components (eg. data table, form, progress indicator) should occupy the full width of the modal. Text, including the paragraph component, should only be 75% of the modal's width.
```
shixiedesign: > Components (eg. data table, form, progress indicator) **should** occupy the full width of the modal. Text, including the paragraph component, should only be 75% of the modal's width.

I will change that to 

> Components (eg. data table, form, progress indicator) **may** occupy up to full width of the modal. Text, including the paragraph component, should only use 75% of the modal's width.

shixiedesign: Also, feels like the width guidance potentially belong to **Body** section? It doesn't apply to header and footer.

src/content/components/modal/usage.mdx
```diff
@@ -3,32 +3,43 @@ title: Modal
...
 
-Include a heading within the modal that mirrors the action or button that was clicked by the user. Headers include a close button “x” in the upper right-hand corner of the modal.
+The header of your modal should mirror the action that launched the modal. Headers must include a close button “✕” in the upper right-hand corner of the modal. You can also include an optional label above your header text. This is an opportunity to offer additional context.
 
 #### Body
 
-The body content within a modal should be as minimal as possible. Components that may be used in modals include: form fields, text area, select, and radio buttons.
+A modal should have minimal body content. Components that may be used in modals include: form fields, text area, select, and radio buttons.
 
 #### Footer
```
shixiedesign: Lemme make some images for this section.

src/content/components/modal/usage.mdx
connor-leech: Thanks!

src/content/components/modal/usage.mdx
connor-leech: Changes made!

src/content/components/modal/usage.mdx
connor-leech: We've talked about making changes, including an acknowledgment for passive modals. We should make an issue and open it up for discussion.

aagonzales: Yeah modals in general need a pass. But thats a good point, transactional probably shouldn't have the "click anywhere to close" function

PR #1711, opened 2019-06-07
Docs: Import pattern

arnaudgillard:
Closes #

{New import pattern}

#### Changelog

**New**

- {New images and text}

**Changed**

- {Navigation}

**Removed**

- {{removed thing}}


lovemecomputer:
playing of of @jnm2377 's comment, but with a different IA thought:

one way to structure this could be, instead of adding each of these pages to the side nav individually, have one side nav item called "File Operations" (or whatever), and then each page for "Download", "Import", and "Delete" could be a tab within!

vikkipaterson:
@lovemecomputer I like that suggestion - I could imagine a file operation patterns section which contained the patterns for 
- import
- export
- upload
- download
Ultimately we are happy to go with what you guys would be happy with so let me know if we should combine them or if you are happy with four separate patterns.

lovemecomputer:
@vikkipaterson that does make sense! there's been a lot of higher-level UX/IA work in our last phase, hopefully we'll start to have a clearer picture of how to merge this kind of guidance into the site soon : )

connor-leech:
Closing due to inactivity. This is blocked by a lack of dedicated visual designers to polish/approve. Will be reopened when picked up.

Ref: https://github.com/carbon-design-system/carbon-website/issues/1859

PR #1714, opened 2019-06-07
docs: Export pattern

arnaudgillard:
Closes #

{New export pattern from hybrid cloud}

#### Changelog

**New**

- {New images and text}

**Changed**

- {Changed navigation}

**Removed**

- {{removed thing}}


aledavila:
@connor-leech final review thanks

aagonzales:
The images seem like they might be using the wrong sizes and aren't exported correctly, they look a little pixelated. 

- Images should be exported at 2x resolution. 
- Images dimensions and component size:
   - for 8 column image component use width: 736px
   - for 12 column image component use width: 1120px
- I think the first two images could use the 12column image component to help see the detail a little better.

arnaudgillard:
Hello! I've just added a new update with larger images and an additional image in the last section.

aagonzales:
Hi thanks for updating! Unfortunately, I'm going to be picky about some things, sorry! 

This image is the right size but the modal in the image itself looks disproportionate to the content behind it. Looks like maybe the background got scaled up but not the modal.

**As-is:**
image
**To-be:**
image


And then the images under "editable name" have been exported at the wrong size and therefore are being scaled down in the image component. If you look at the [modal](https://www.carbondesignsystem.com/components/modal/usage#format) usage page you can see more what the scale of them should be. To change this you just need to make the artboard for this image at 736px (looks like its currently at 1120px width) then it will be scaled correctly inside the component.

**As-is:**
image
**To-be:**
image

arnaudgillard:
@aagonzales I've just updated the imagery. The sizing should hopefully be correct. I have also changed the "Exporting..." status to the one recommended in the "deleting pattern". This should follow the carbon requirements more closely. 

Let me know if there is anything else to change!

PR #1716, opened 2019-06-07
chore: update to 10.3

### LINKED ISSUE #1639: [Button, website] Update Button page with icon buttons
Update website component Button page with newly merged **Icon buttons** .
### Add these to website's Button component example: 
- Primary icon button (default & small)
- Secondary icon button (default & small)

image
image

image
image

alisonjoseph:
- Update to 10.3 and 7.3 for component package
- Update left nav override styles to be compatible with new shell updates
- fixes some console errors

closes #1639  

Updating from `7.2` to `7.3` for `carbon-components-react` is breaking the Gatsby build. This is related to the tab/tabs component but beyond narrowing it down to that I'm not sure how to fix. 

This is the build error https://reactjs.org/docs/error-decoder.html/?invariant=130&args%5B%5D=undefined
> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.[missing argument]

Any ideas on what might have changed with the React library from `7.2` to `7.3` ?


alisonjoseph:
Updating from `7.2` to `7.3` for `carbon-components-react` is breaking the Gatsby build. This is related to the tab/tabs component but beyond narrowing it down to that I'm not sure how to fix. 

This is the build error https://reactjs.org/docs/error-decoder.html/?invariant=130&args%5B%5D=undefined
> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.[missing argument]

Any ideas on what might have changed with the React library from `7.2` to `7.3` ?

tay1orjones:
Could this be cut into a new alpha release? Our addons site depends on this as it's not yet on the new gatsby theme. 

We're attempting to update our `breakpoint` mixin usage to have the prefix. Our build is failing without the `breakpoint` fix to `_typeface-sub-families.scss`.

alisonjoseph:
Just published `0.0.1-alpha.24` @tay1orjones, hopefully, that works ok for you guys and doesn't break stuff. 🙏 Let me know if you want help converting your site to use the theme. We're in the middle of moving the Design Language site and Carbon to use it right now. 😅

PR #1717, opened 2019-06-07
docs(tutorial): update and publish step 3

mattrosno:
DO NOT MERGE

We'll do a timed merge to control when step 3 is published.


src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -301,233 +266,188 @@ const getRowItems = rows =>
...
-    if (loading)
-      return (
-        <DataTableSkeleton
-          columnCount={headers.length + 1}
-          rowCount={12}
-          headers={headers}
-        />
-      );
+    // Wait for the request to complete
+    if (loading) return <>Loading...</>;
```
vpicone: return "Loading..."

src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -301,233 +266,188 @@ const getRowItems = rows =>
...
   }}
 </Query>
 ```
 
-We're not rendering the table yet, but if you refresh the page, you'll see the data table skeleton while the data loads! Also, if you view your browser's network requests, you can see XHR request response to see what's getting returned to the browser.
-
-_Note: We only pass the rows that we want our table to display, we can do this by slicing the our array of rows depending on the first item and the page size._
+The page will look the same as we're still rendering our static example rows, but if you view your browser's console (e.g. [Chrome DevTools](https://developers.google.com/web/tools/chrome-devtools/)), you should see the response from GitHub!
 
-_Note: The Pagination component isn’t inherently connected in any way to the DataTable, we need to tell it what to do when a change occurs using the onChange prop. This includes both page size changes and displaying different rows._.
```
vpicone: We need this.

src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -301,233 +266,188 @@ const getRowItems = rows =>
...
+const getRowDescription = rowId => {
+  const row = rows.find(({ id }) => id === rowId);
+  return row ? row.description : '';
+};
 ```
 
-### Component cleanup
+Finally, in `RepoTable.js`, replace `<p>Row description</p>` with:
 
-The data table component and its pieces use a common React pattern called [render props](https://reactjs.org/docs/render-props.html). This a really powerful way for libraries to give developers control of rendering and manipulating their data.
```
vpicone: This is critical information and the biggest source of issues in slack for DataTable

src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -301,233 +266,188 @@ const getRowItems = rows =>
...
+
+##### src/content/RepoPage/RepoPage.js
+
+```javascript
+const [totalItems, setTotalItems] = useState(0);
+const [firstRowIndex, setFirstRowIndex] = useState(0);
+const [currentPageSize, setCurrentPageSize] = useState(10);
 ```
 
-_Note: The `getTableProps`, `getHeaderProps`, and `getRowProps` functions each tell the DataTable component to process and return the necessary props that are needed by the Table, Header and Row components respectively. We take the return of that function (an object containing the props) and [spread it out](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes) into the component._.
```
vpicone: This is also critical info was it moved?

vpicone:
1) Libraries overview + graphQL query writing
2) Loading, error, and normal states
3) Necessary components (Skeleton, DataTable, Pagination)
3) Common DataTable miss-steps

src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -131,74 +132,53 @@ const client = new ApolloClient({
...
-  TableBody,
-  TableExpandRow,
-  TableCell,
-  TableExpandedRow,
-} from 'carbon-components-react';
-```
-
 ### Query
 
 Next we'll assemble our GraphQL query to fetch only the data we need from the GraphQL API. We'll do this using the `gql` helper we just imported. The `gql` helper lets you write GraphQL queries using interpolated strings (backticks) in JavaScript. In addition, we'll be using the `Query` component from `react-apollo` which gives us some great information about our query's loading state in addition to the data.
```
vpicone:
```suggestion

You can use GitHub's [explorer](https://developer.github.com/v4/explorer/) tool to write and test your own queries. Try copying the query below and experiment with changing the properties. You can also click the "Docs" button in the top right of the explorer to view all of the available data and query parameters.

```

src/content/tutorial/react-step-3/react-step-3.mdx
mattrosno: It should have been. I've made sure this is included when data table is first used in the previous step.

src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -301,233 +268,200 @@ const getRowItems = rows =>
...
       </>
     );
   }}
 </Query>
 ```
 
-We're not rendering the table yet, but if you refresh the page, you'll see the data table skeleton while the data loads! Also, if you view your browser's network requests, you can see XHR request response to see what's getting returned to the browser.
+The page will look the same as we're still rendering our static example rows, but if you view your browser's console (e.g. [Chrome DevTools](https://developers.google.com/web/tools/chrome-devtools/)), you should see the response from GitHub!
 
-_Note: We only pass the rows that we want our table to display, we can do this by slicing the our array of rows depending on the first item and the page size._
```
vpicone: Did we add this bit somewhere?

src/content/tutorial/react-step-3/react-step-3.mdx
```diff
@@ -301,233 +268,200 @@ const getRowItems = rows =>
...
+
+Revisiting `RepoTable.js`, we are already passing in our row objects along with headers for each column. The `render` prop is being used to tell the data table how to render the headers and rows. That prop takes a function that receives the processed headers and rows as arguments as well as some helper functions for rendering the table.
+
+One common hurdle with the data table is how to access data that might be not correspond with a table column but is needed to compute the value of a cell that does. The data table component processes and controls only the row properties which corresponds to headers (columns). Because of this, the `rows` object you get access to in the render prop function is _different_ than the one you passed in to the `rows` prop.
 
-Import the various components that we'll need to render our table at the top of `RepoTable.js`.
+We need to modify one apsect of the data table because if you expand a row, it says `Row description`. We want to update that with the repository description coming from the GitHub API. This is an example of where we need a simple look-up function to find the data we care about - data that does not directly correspond to a column.
+
```
vpicone: I think this rewrite does no longer makes it seem like this change solves the previous problem outlined in the previous component. It’s also jarring hopping to the other file without context.

src/content/tutorial/react-step-3/react-step-3.mdx
mattrosno: Added.

src/content/tutorial/react-step-3/react-step-3.mdx
mattrosno: Now that we have GH data, we need to make one change to RepoTable.js to update "Row description". I attempted to set context on why we need to touch that file. Maybe updating heading from "Data table data" to "Render repository descriptions" will help.

PR #1724, opened 2019-06-10
feat: add 10.3 shell component examples

jnm2377:
- [x] add code page for UI Shell header
- [x] add UI Shell left panel
- [x] add UI Shell right panel
- [x] fix header styles
- [x] fix sidenav styles
- [x] fix right panel styles
- [x] update component status page
- [x] add component readme
- [x] fix component readme?

https://deploy-preview-1724--loving-kepler-d07227.netlify.com/components/UI-shell-header/code


jnm2377:
Question: I've added the overriding styles for left side nav, but the `Fixed left panel` example for some reason seems to still have the un-overridden styles on hover. When I inspect it though, it has the right styles, but for some reason I'm not actually seeing that. Is that happening for y'all too? @alisonjoseph @designertyler 
Screen Shot 2019-06-12 at 10 54 27 AM

jnm2377:
Also, I added the ui-shell component readme just like we do for other component pages but for some reason the styles look awful. Wondering if it's something I should override on the website or if it's something wrong with the readme itself? @alisonjoseph 

The read me, for reference: https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/ui-shell/README.md

alisonjoseph:
I think the readme stuff has to do with some different formatting in this readme file that we haven't seen before. (blockquote, lists, h1, h2 + h4) Let me fix and add to your PR

jnm2377:
ok, @designertyler  tried @alisonjoseph's idea of embedding an iframe for component examples instead. This may actually be nicer anyway since people will be able to see code with it like our vanilla components. Thoughts?

Also added the styles Alison sent me to fix the readme 👍 ✅

PR #1741, opened 2019-06-13
fix(markdown): remove unused classname prop in `a` component

### LINKED ISSUE #1740: Links generated from markdown have `undefined` className
## Detailed description

Anchor elements are expecting `this.props.className` in the `markdown` component: https://github.com/carbon-design-system/carbon-website/blob/master/src/components/markdown/Markdown.js#L177

```javascript
export class a extends React.Component {
  render() {
    return (
      <a {...this.props} className={`${this.props.className} bx--link`}>
        {this.props.children}
      </a>
    );
  }
}
```

But they never get that class because these links are rendered from Markdown files that don't include any extra classes in the link syntax. As a result, all links inside main page content as `undefined` class.

Example here (inspect any link inside the main page content, generated from Markdown files -- including inside `AnchorLinks`): https://www.carbondesignsystem.com/getting-started/developers/vanilla

jendowns:
Closes #1740 

#### Changelog

**Removed**

- remove unused `this.props.classname`


jendowns:
Could someone restart the deploy preview for me? 😇 Looks like it timed out

Issue #1787, opened 2019-06-26
<IconLibrary> component - cupcake experience implementation


lovemecomputer:
ref https://github.com/carbon-design-system/carbon-website/issues/1485
sketch file: https://ibm.box.com/s/ojsfg916c8rc2hsntbvjrj836iesas2p

---

this issue is to track implementation of `<IconLibrary>` component as specc'd in #1485 . 

- [x] presence of search bar component
- [x] headers for each icon size
- [x] icon card
  - [x] icon name ("friendly name"?)
  - [x] icon asset
  - [ ] ~functionality when clicking or hovering on icon cards?~ not in cupcake experience
- [x] Block/card: "No results found in this size"
- [x] "No results found" display: CTA "submit an icon" button needs to link to new monorepo package
- [x] tabbing order: currently, hitting tab from the search field takes you straight to the "edit this page" link at the bottom of the page
- [x] "x results found" message under search bar

--- 
**deferred/additional items**
- [ ] fuzzy search implementation to be split into separate PR
- [ ] additional `<IconLibrary>` enhancements covered in full spec: https://github.com/carbon-design-system/carbon-website/issues/1094

Issue #1485, opened 2019-04-11
Icon page cupcake experience 


designertyler:
Cupcake (4/25)
"As a product designer or developer I need to..."
- [x] see all icons so that I can know the icon I need exists
https://github.com/carbon-design-system/carbon-website/issues/1325

- [x] see the icons when I arrive on the icon page so that I know I'm in the right spot.  (smaller resource cards)
https://github.com/carbon-design-system/issue-tracking/issues/107

- [ ] link to the Sketch cloud library so that I can see all of the library while working in Sketch.
https://github.com/IBM/carbon-design-kit/issues/89

- [x] updated "no results" page https://github.com/carbon-design-system/carbon-website/issues/866

- [x] find the icon repository so that I... [specify goal].
- [x] navigate to the icon repo so that I can download the icons directly to make edits and contribute back.
- [x] can get automatic updates when the language/system team makes changes so the icons I use will always be up to date.

designertyler:
## Cupcake Designs

### Default state
![Icon_Library_all](https://user-images.githubusercontent.com/19270502/56527949-12c28500-6514-11e9-9bfa-2d57f54933cc.png)

### Search results
![Icon_Library filtered](https://user-images.githubusercontent.com/19270502/56527972-1ce48380-6514-11e9-96be-0211ee753413.png)

### No results
![Icon_Library no results](https://user-images.githubusercontent.com/19270502/56527996-24a42800-6514-11e9-8bdb-c2217be49657.png)


### Spacing
![Icon_Library filtered_specs](https://user-images.githubusercontent.com/19270502/56528213-72b92b80-6514-11e9-8c7f-6cc73f474a6b.png)
![Icon_Library no results_specs](https://user-images.githubusercontent.com/19270502/56528214-72b92b80-6514-11e9-8e26-843e07c435f5.png)
![Icon_Library_all_specs](https://user-images.githubusercontent.com/19270502/56528215-72b92b80-6514-11e9-952e-8b252248a96b.png)

chrisconnors-ibm:
get bent stale[bot]

chrisconnors-ibm:
still active.

joshblack:
![image](https://user-images.githubusercontent.com/3901764/59875583-bab6cc00-9366-11e9-92e3-0a06e9023fda.png)

Would this tile need to be renamed to our kit?

joshblack:
[...]

PR #1789, opened 2019-06-26
icon library pages cupcake updates

lovemecomputer:
closes https://github.com/carbon-design-system/carbon-website/issues/1787

https://deploy-preview-1789--carbon-website.netlify.com/guidelines/iconography/usage
https://deploy-preview-1789--carbon-website.netlify.com/guidelines/iconography/library

---

**Changes:**
- include block for empty size categories
- update no-results link
- change icons to be `<li>`s inside a `<ul>`
- add counter for search results
- **Usage** page is now first tab
- resource links removed from **Library** page and put into **Usage** page
- refactor icon list flexbox layout code to resolve some small issues/create something more easily maintainable/supportable
- remove glyphs from displayed icon data

**Testing notes:**
- navigate to page **Guidelines > Iconography**. see how it looks, and if the current basic search functionality works for you!

---

**Deferred items**
- " link to the Sketch cloud library so that I can see all of the library while working in Sketch.
carbon-design-system/carbon-design-kit#89" — icons in sketch kit still WIP, currently still linking to ibm design language library


src/components/IconLibrary/IconLibrary.js
```diff
@@ -212,32 +241,64 @@ function createIconSections(icons, filteredIcons) {
...
     ));
 }
 
+/**
+ * render list of available icons for this category.
+ * if none are available, return a blank block
+ */
+
+function renderIconList(categoryArray, filteredList) {
+  let displayedIconsList = [];
```
joshblack:
```suggestion
  const displayedIconList = categoryArray.filter(icon => filteredList.indexOf(icon.name) !== -1);
```

src/components/IconLibrary/IconLibrary.js
```diff
@@ -212,32 +241,64 @@ function createIconSections(icons, filteredIcons) {
...
+  categoryArray.forEach(icon => {
+    if (filteredList.indexOf(icon.name) !== -1) {
+      displayedIconsList.push(icon);
+    }
+  });
+  const displayedIcons = displayedIconsList.map(renderIcon);
+  return displayedIcons.length > 0 ? (
```
joshblack: Might help for readability, definitely personal preference:
```suggestion
  if (displayedIcons.length > 0) {
    return displayedIconsList.map(renderIcon);
  }
  
  return (
    <li className="icon">
      <div className="icon__card" />
      <h5 className="icon__card-title" title="No results in this size.">
        No results in this size.
      </h5>
    </li>
  );
```

src/components/IconLibrary/IconLibrary.js
```diff
@@ -212,32 +241,64 @@ function createIconSections(icons, filteredIcons) {
       return groups[size].length !== 0;
     })
     .map(size => (
-      <section key={size} className="icon-size">
+      <section
+        key={size}
+        className="icon-size"
+        aria-labelledby={`icon-h2 icon-h2-${size}`}>
```
joshblack: Would this work? Had always assumed `aria-labelledyby` only worked on ids.

src/components/IconLibrary/IconLibrary.js
```diff
@@ -212,32 +241,64 @@ function createIconSections(icons, filteredIcons) {
...
-          <h2 className="icon-h2">{isNaN(size) ? size : `${size}x${size}`}</h2>
+          <h2 className={`icon-h2 icon-h2-${size}`}>
+            {isNaN(size) ? size : `${size}x${size}`}
+          </h2>
         </header>
-        <div className="icon-container">
-          {groups[size]
+        <ul className="icon-container" tabindex="0">
+          {renderIconList(groups[size], filteredIcons)}
+          {/* {groups[size]
```
joshblack: Do we need this commented block anymore?

src/components/IconLibrary/IconLibrary.js
```diff
@@ -68,6 +68,22 @@ export default class IconLibrary extends React.Component {
     );
   };
 
+  /**
+   * clear search state to reset view to default
+   */
+  handleClearSearch = event => {
+    event.preventDefault();
+    const searchValue = '';
```
joshblack:
```suggestion
```

src/components/IconLibrary/IconLibrary.js
```diff
@@ -68,6 +68,22 @@ export default class IconLibrary extends React.Component {
...
+    this.setState(
+      {
+        searchValue,
```
joshblack:
```suggestion
        searchValue: '',
```

src/components/IconLibrary/icon-library.scss
```diff
@@ -0,0 +1,10 @@
+.icon-search--wrapper {
+  position: relative;
+}
+
+.icon-search--status {
+  position: absolute;
+  top: 3.6rem;
```
joshblack: Could you leave a comment on positioning and why this `3.6rem` is needed? 🙏

src/components/IconLibrary/icon-library.scss
```diff
@@ -0,0 +1,10 @@
...
+  padding-left: $spacing-05;
+  @include carbon--type-style('body-short-02');
```
joshblack: Style nit: mixins tend to be placed at the top of blocks

joshblack:
One note for CI, you'll probably need to do:

```
yarn clean
yarn cache clean
yarn install --frozen-lockfile
```

And then the new dependencies will pop up 😄

joshblack:
Random question, should the `ul` have a tabindex?

src/content/guidelines/iconography/library.mdx
```diff
@@ -6,9 +6,8 @@ tabs: ['Library', 'Usage', 'Contribute']
 
 import IconLibrary from '../../../../src/components/IconLibrary';
 
-
 <Row className="tile--group">
-<Column offsetLg="4" colMd="4" colLg="4" noGutterSm>
+<Column offsetLg="4" colSm="3" colMd="4" colLg="4" noGutterSm>
```
alisonjoseph:
```suggestion
<Column offsetLg="4" colSm="4" colMd="4" colLg="3" noGutterSm>
```

I believe this is the correct sizing for the cards, however, definitely want to check w/ design. Full width at mobile, 1/2 at tablet, and then they want smaller than normal at the larger breakpoint, so only 3 cols instead of the normal 4.

src/content/guidelines/iconography/library.mdx
```diff
@@ -18,7 +17,7 @@ import IconLibrary from '../../../../src/components/IconLibrary';
 
  </ClickableTile> 
 </Column>
-<Column colLg="4" noGutterSm>
+<Column colSm="3" colMd="4" colLg="4" noGutterSm>
```
alisonjoseph:
```suggestion
<Column colSm="4" colMd="4" colLg="3" noGutterSm>
```

src/content/guidelines/iconography/library.mdx
lovemecomputer: thanks alison! i experimented with different settings to try to find the best behavior. i'll look to your previous work as an example to follow!

lovemecomputer: also hmm in #1485 a specific request is "see the icons when I arrive on the icon page so that I know I'm in the right spot.  (smaller resource cards)", so i'll check with tyler : )

lovemecomputer: there was a lot of unsolved design discussion, so for now to match the IDL site, all resource links have been moved to usage page

src/content/guidelines/iconography/usage.mdx
```diff
@@ -29,6 +29,16 @@ tabs: ['Library', 'Usage', 'Contribute']
 
  </ClickableTile>
 </Column>
+<Column colLg="4" colMd="4"  noGutterSm>
```
alisonjoseph:
```suggestion
<Column offsetLg="4" colLg="4" colMd="4"  noGutterSm>
```

This tile needs the offset class to appear below the other two

src/components/IconCard/icon-card.scss
```diff
@@ -1,22 +1,123 @@
...
+  }
+  &:nth-child(n):nth-child(-n + 2) {
+    border-top: none;
+  }
+  &:nth-child(2n - 1) {
+    border-left: none;
+  }
+}
+
+@media all and (min-width: 515px) {
```
alisonjoseph: Is there a reason we aren't using the Carbon media queries? can you add a comment saying why or if possible switch to the Carbon defaults?

src/components/IconLibrary/IconLibrary.js
```diff
@@ -212,32 +244,63 @@ function createIconSections(icons, filteredIcons) {
...
  * Renders an individual icon
+ * passing `null` renders no-results icon block
  */
 function renderIcon(icon) {
   return (
-    <div key={icon.name} className="icon">
-      <div className="icon__card">
-        <icon.Component />
+    <li key={icon.name} className="icon__container">
+      <div className="icon__aspect-ratio">
```
alisonjoseph: We could use the built in aspect ratio class names here if you want? https://carbon-elements.netlify.com/grid/examples/preview/#aspect-ratio

src/components/IconCard/icon-card.scss
lovemecomputer: the component did have bespoke breakpoints before, and these are modified but based on the same reasoning. 

the breakpoints are hard coded and specific values to keep the icon tiles fitting properly (filling the full with of the container) at every possible window size

chrisconnors-ibm:
@designertyler is the spec creator and 1st reviewer for "As specified"

src/components/IconLibrary/IconLibrary.js
lovemecomputer: oh yeah i should've used those instead of re-writing 😅

src/content/guidelines/iconography/usage.mdx
lovemecomputer: cool thanks, comitted ur suggestion! i was initially assuming it'd just be 3 flowing across but this is better + designer-approved now : )

chrisconnors-ibm:
We are getting so close! Couple of requests: 
please ignore Glyphs* (found @ `carbon/packages/⁨icons/⁨src/⁨svg⁩`) and there shouldn't be undefined categories
image

*sketch plugin should ignore them too

chrisconnors-ibm:
(if any reviewers want to check the aliases are working, they're in `metadata.yml`)

lovemecomputer:
@chrisconnors-ibm this PR only updates the IconLibarary component's front-end, no the search algorithm. thus, aliases are not supported while previewing this PR

chrisconnors-ibm:
I just searched for:
plus
positive
addition
increase

and all returned "no results" rather than `add`

(these are the first set of aliases in `metadata.yml`, L8-L12)

chrisconnors-ibm:
> @chrisconnors-ibm this PR only updates the IconLibarary component's front-end, no the search algorithm. thus, aliases are not supported while previewing this PR

where do we test that?

lovemecomputer:
@chrisconnors-ibm i do not know what has been scoped or worked on on that front, @joshblack may be able to answer ur questions. afaik, for the next round of work ("birthday cake"?), that is when we're planning to implement support for category info, and i assume it's the update where we would expose the rest of the metadata like aliases as well.

designertyler:
Yes, icon alias was meant for birthday cake.

The "No results in this size." tiles are missing the second bit of info that says scale up the 16px to get the 20px or scale down the 32px to get the 24px. We can modify the text to something like "Scale the closest size to use this icon" if the metadata doesn't allow us to know when to scale up or down.

image

lovemecomputer:
thanks, @designertyler ! for the moment i pushed up an updated with our suggested text for the "no results found" tile. will review tomorrow if feasible to include the the size suggestions or if should be implemented in the birthday cake version (since that's when we're introducing the new metadata to the component)

lovemecomputer:
also fixed the border on the no-results screen button you showed me!

@designertyler one question: what about the edge state where, when you search for a term (e.g. `caret`) which has a glyph designed, but is not included in the displayed icon sizes? if you search for `caret` in the current implementation, you can see that each size category renders the "no results in this size" card… 

so, that's a state where our site has no results displayed… we can still display the "no results/submit an icon" CTA, but i wasn't sure if it was correct to provide a "submit an icon" CTA when the asset exists in in the package other (`Glyph`) forms…

chrisconnors-ibm:
>which has a glyph designed, but is not included in the displayed icon sizes? if you search for caret in the current implementation, you can see that each size category renders the "no results in this size" card…

glyphs (found @ carbon/packages/⁨icons/⁨src/⁨svg⁩) should be ignored for results. they have (or should have) 16x16 equivalents. `caret--*.svg` exists in `/svg/16/`

src/components/IconCard/icon-card.scss
```diff
@@ -1,27 +1,119 @@
...
+    &:nth-child(n):nth-child(-n + 3) {
+      border-top: none;
+    }
+    &:nth-child(3n - 2) {
+      border-left: none;
+    }
+  }
+}
+
+@media all and (min-width: 730px) {
```
joshblack: The styles and media queries here seem very specific and potentially fragile with respect to minor tweaks. Would we want to explore another way of achieving this layout? Not quite sure what the rules are but would be worried about including one-off breakpoints for this. 

Is the sketch kit itself aligned to the grid? Maybe we could use the column classes for this? If not, I wonder if we could use something like CSS Grid to help us out here 🤔 cc @alisonjoseph is that something we can use on the site?

alisonjoseph: We can't use the column classes easily here until we switch this site over to the Gatsby theme / get rid of the offset for each row. (which I already did over on the IDL site https://github.com/carbon-design-system/design-language-website/blob/master/src/components/IconLibrary/icon-library.scss)

I think we can leave it for now, and hopefully switch it when we do the birthday cake version of this site.

lovemecomputer: this is a refactoring of previously-existing CSS, so it isn't a departure from what's live on the current site. the only change is how sizing & flex works for each displayed icon, so that they visually align to edges of their container (instead of being constrained to a maximum width). 

and yeah as alison said we can totally refactor this when merged with the new gatsby theme  :)

joshblack: Sounds good!

lovemecomputer:
@joshblack regarding the disappearing borders, that was an issue with the previous CSS that was resolved in the refactor. maybe you're looking at an old netlify build somehow?

alisonjoseph:
I think this no-results view still needs to get updated.

Screen Shot 2019-07-11 at 11 26 55 AM

designertyler:
Alison's screenshot is cropping off the resource cards that lead someone to submitting a new icon. https://user-images.githubusercontent.com/19270502/56744049-6663e700-673d-11e9-89bc-36aee61dd2e4.png

image

chrisconnors-ibm:
netifly artifact not currently reflecting contents of `elements`; `chev` yields no results (missing `chevron--*.svg`); 16x16 icons = 53 items but there are 114 svgs in `carbon/packages/⁨icons/⁨src/⁨svg⁩/16`

lovemecomputer:
@alisonjoseph thanks alison! that should be updated in the latest commit. i think it's ready!!

lovemecomputer:
@joshblack got it thanks!!

PR #1807, opened 2019-07-08
Adding experimental nav pattern

designertyler:
Closes https://github.com/carbon-design-system/carbon/issues/3001


chrisconnors-ibm:
@aagonzales, @jeanservaas, @janchild @connor-leech your feedback pls

jeanservaas:
Hey tyler! just a couple things"

- I'd make "terms" an H2 in order to be consistent with our hierarchy across the site. 

image

- Also, I'll defer to @janchild and @connor-leech on this, but it seems like we could shorten "Importance to a user's task" to "Importance to task" or something.

- Links in those descriptions are 16px but the text is 14px

image

- This is super nit picky, but it caught my eye -- wondering what happened to the color on those browser action circles (instead of red, yellow and green they look a little sea sick); and then in the slash design example they're a slightly different color than the data table browser... we should streamline that.

image

image

jeanservaas:
I know you hate the anchor links, but when "Terms" becomes an H2, we're gonna have to add it to the anchor links to be consistent with the behavior for the site.

PR #1809, opened 2019-07-08
Update glossary.js to remove `cold-blooded cyborg`

jeoffw:


janchild:
@alisonjoseph, I just reviewed the glossary with @connor-leech and we agreed we should remove all of the idioms. 

We have it on our backlog list to review the nature of this page too. It would be more helpful as a preferred word list (use this versus that) rather than a glossary.

PR #1818, opened 2019-07-10
chore: update to 10.4.0

alisonjoseph:


alisonjoseph:
seeing this bug now with rc 1, weird overlay when you hover on/off pages


website-hover

package.json
```diff
@@ -40,18 +40,13 @@
...
-    "@carbon/layout": "v10.1.1",
-    "@carbon/type": "v10.1.1",
+    "@carbon/elements": "10.4.0",
     "@mdx-js/mdx": "^1.0.0",
     "@mdx-js/react": "^1.0.0",
     "@reach/router": "^1.2.1",
     "@vimeo/player": "^2.6.7",
-    "carbon-components": "v10.3.0",
-    "carbon-components-react": "v7.3.0",
+    "carbon-components": "10.4.0",
```
joshblack:
```suggestion
    "carbon-components": "10.4.1",
```

package.json
```diff
@@ -40,18 +40,13 @@
...
+    "carbon-components-react": "7.4.0",
```
joshblack:
```suggestion
    "carbon-components-react": "7.4.1",
```

package.json
```diff
@@ -40,18 +40,13 @@
   },
   "dependencies": {
     "@carbon/addons-website": "^0.0.1-alpha.23",
-    "@carbon/icons-react": "10.4.0-alpha.5390",
-    "@carbon/colors": "v10.1.1",
-    "@carbon/grid": "v10.1.1",
-    "@carbon/import-once": "v10.1.0",
-    "@carbon/layout": "v10.1.1",
-    "@carbon/type": "v10.1.1",
+    "@carbon/elements": "10.4.0",
```
joshblack: We might need to keep `@carbon/icons-react` as a dependency since it's not included in `@carbon/elements` by default.

alisonjoseph: Oh, didn't realize. I'll update!

alisonjoseph:
@emyarod colors, grid, layout, type are all included with @carbon/elements. Just updated to include that and not all of them individually.

PR #1886, opened 2019-08-16
Correct usage.mdx

haydaralrikabi:
The tooltip component cannot be aligned to (start/center/end). This sort of alignment is only available for the definition tooltip.


connor-leech:
Closing due to lack of activity

